PR #56, opened 2024-06-08
fix: sort navigation items numerically in `objectToSortedArray` function

morinokami:
I discovered a bug where parts, chapters, and lessons in navList were being sorted by comparing numbers as strings, causing issues like Chapter 10 being placed before Chapter 2.

Reproduction: https://stackblitz.com/~/github.com/morinokami/tutorialkit-sort-problem

I have fixed this by comparing each key as a numeric value in the objectToSortedArray function.

### DIFF
--- a/.changeset/smooth-camels-fail.md
+++ b/.changeset/smooth-camels-fail.md
@@ -1,0 +1,5 @@
+---
+"@tutorialkit/astro": patch
+---
+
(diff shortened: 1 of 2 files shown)


d3lm:
Thanks for the PR 🙏 Good catch!

SamVerschueren (approved):

PR #57, opened 2024-06-08
feat: set default package manager to the one used with `create` command

morinokami:
Currently, when creating a project using a command like `npm create tutorial`, the default package manager for installing dependencies is [hardcoded to pnpm](https://github.com/stackblitz/tutorialkit/blob/94546823c3393e4d3005ff78038119692475466b/packages/cli/src/commands/create/dependencies.ts#L81). However, it seems more intuitive to use the package manager that was used to run the create command as the default. This PR makes that change. Additionally, if the package manager cannot be determined, the fallback default is set to npm.

### DIFF
--- a/packages/cli/src/commands/create/dependencies.ts
+++ b/packages/cli/src/commands/create/dependencies.ts
@@ -77,6 +77,7 @@ async function getPackageManager() {
   const installedPackageManagers = await getInstalledPackageManagers();
 
   let initialValue = process.env.npm_config_user_agent?.split('/')[0] as PackageManager | undefined;
+
   if (!installedPackageManagers.includes(initialValue)) {
     initialValue = 'npm';
   }
--- a/packages/cli/src/commands/create/dependencies.ts
+++ b/packages/cli/src/commands/create/dependencies.ts
@@ -76,7 +76,10 @@ export async function installDependencies(cwd: string, flags: CreateOptions) {
 async function getPackageManager() {
   const installedPackageManagers = await getInstalledPackageManagers();
 
-  const initialValue = process.env.npm_config_user_agent?.split('/')[0] ?? 'npm';
+  let initialValue = process.env.npm_config_user_agent?.split('/')[0] as PackageManager | undefined;
+  if (!installedPackageManagers.includes(initialValue)) {
+    initialValue = 'npm';
+  }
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/create/dependencies.ts
```diff
@@ -76,9 +76,11 @@ export async function installDependencies(cwd: string, flags: CreateOptions) {
 async function getPackageManager() {
   const installedPackageManagers = await getInstalledPackageManagers();
 
+  const initialValue = process.env.npm_config_user_agent?.split('/')[0] ?? 'npm';
```
SamVerschueren, line 79: I think it would be good if we double-checked to make sure it's part of the installed package managers?

```ts
  let initialValue = process.env.npm_config_user_agent?.split('/')[0] as PackageManager | undefined;

  if (!installedPackageManagers.includes(initialValue)) {
    initialValue = 'npm';
  }
```

Nemikolh: I agree, either that or we assume that the value provided will always work (given it succeeded already) and we make sure that `installPackageManagers` includes it.

Nemikolh (approved):
Wow thanks a lot for the PR @morinokami ! :star_struck:

This looks really good! :clap:

packages/cli/src/commands/create/dependencies.ts
morinokami: @SamVerschueren @Nemikolh Thank you for your reviews! I've updated the code to match Sam's suggestion!

packages/cli/src/commands/create/dependencies.ts
```diff
@@ -76,9 +76,14 @@ export async function installDependencies(cwd: string, flags: CreateOptions) {
 async function getPackageManager() {
   const installedPackageManagers = await getInstalledPackageManagers();
 
+  let initialValue = process.env.npm_config_user_agent?.split('/')[0] as PackageManager | undefined;
+  if (!installedPackageManagers.includes(initialValue)) {
```
SamVerschueren, line 80: nitpick

```suggestion

  if (!installedPackageManagers.includes(initialValue)) {
```

SamVerschueren (approved):
Awesome work @morinokami 🙏 !

PR #59, opened 2024-06-10
refactor: move runtime logic in TutorialStore

Nemikolh:
This PR refactors the logic that lived in `WorkspacePanel` to be in `@tutorialkit/runtime`. The idea behind this move is to support:
* Custom components that import a virtual module like `tutorialkit:state` to access the `tutorialStore`. This would then let them watch some state or modify it such as: run a command in a terminal, navigate a preview to a path, etc...
* Support components written for a different framework like svelte, etc...

### DIFF
--- a/packages/components/react/src/CodeMirrorEditor/index.tsx
+++ b/packages/components/react/src/CodeMirrorEditor/index.tsx
@@ -14,7 +14,7 @@ import {
   scrollPastEnd,
 } from '@codemirror/view';
 import { useEffect, useRef, useState, type MutableRefObject } from 'react';
-import type { Theme } from '../types';
+import type { Theme } from '../types.js';
 import { debounce } from '../utils/debounce.js';
 import { BinaryContent } from './BinaryContent.js';
 import { getTheme, reconfigureTheme } from './cm-theme.js';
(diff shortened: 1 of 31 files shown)


packages/components/react/src/CodeMirrorEditor/index.tsx
```diff
@@ -46,7 +46,7 @@ export type OnChangeCallback = (update: EditorUpdate) => void;
 export type OnScrollCallback = (position: ScrollPosition) => void;
 
 interface Props {
-  reset?: unknown;
+  id?: unknown;
```
Nemikolh, line 49: Renamed to be consistent with how `'react-resizable-panels'` uses their `id` (given that here it has the same purpose).

packages/components/react/src/Panels/EditorPanel.tsx
```diff
@@ -9,48 +9,42 @@ import {
...
   editorDocument,
+  selectedFile,
   onEditorChange,
   onEditorScroll,
   onHelpClick,
-  onFileClick,
+  onFileSelect,
 }: Props) {
   const fileTreePanelRef = useRef<ImperativePanelHandle>(null);
-  const [selectedFile, setSelectedFile] = useState(lesson.data.focus);
```
Nemikolh, line 40: This state was lifted up in the `TutorialStore`.

SamVerschueren (approved):
Awesome refactor. This really improves the `@tutorialkit/runtime`!

PR #60, opened 2024-06-10
test(cli): add test for creating and building a project

### LINKED ISSUE #54: Console errors on npm run dev
### Describe the bug

image

_No response_

### Steps to reproduce

npm i
npm run dev

### Expected behavior

no errors

### Screenshots

_No response_

### Platform

Windows, Chrome. Node 20, npm 10.7.0

### Additional context

_No response_

SamVerschueren:
fixes #54

The idea here is that CI also creates the template (which gets generated with `npm create tutorial`) and builds it. This way we can verify that we don't break the template if we update dependencies.

The PR also tests on Windows and addresses some issues with globs that didn't pick up the correct files.


Nemikolh (approved):
Yay!! Amazing :star_struck: 

This should close #54

.github/workflows/ci.yaml
```diff
@@ -7,8 +7,21 @@ on:
 jobs:
   test:
```
d3lm, line 8: Unrelated but should we rename this since it's not just tests?

.github/workflows/ci.yaml
```diff
@@ -7,8 +7,21 @@ on:
...
     name: Test
```
d3lm, line 9: Same for the name. WDYT?

Nemikolh:
@d3lm Hey! Do you think you could update the required status checks to match what's in this PR?

packages/cli/tests/create-tutorial.test.ts
```diff
@@ -0,0 +1,59 @@
...
+beforeAll(async () => {
+  await fs.rm(tmpDir, { force: true, recursive: true });
+  await fs.mkdir(tmpDir);
+});
+
+afterAll(async () => {
+  await fs.rm(tmpDir, { force: true, recursive: true });
+});
+
+test('test creating a project', async () => {
```
d3lm, line 18:
```suggestion
test('create a project', async () => {
```

packages/cli/tests/create-tutorial.test.ts
```diff
@@ -0,0 +1,59 @@
...
+      TK_DIRECTORY: path.resolve(cwd, '..'),
+    },
+  });
+
+  const projectFiles = await fs.readdir(dest, { recursive: true });
+
+  expect(projectFiles.map(normaliseSlash).sort()).toMatchSnapshot();
+});
+
+test('test creating and building a project', async () => {
```
d3lm, line 34:
```suggestion
test('create and build a project', async () => {
```

d3lm:
> Do you think you could update the required status checks to match what's in this PR?

Yep for sure!

.github/workflows/ci.yaml
SamVerschueren: I still feel it's mainly testing related. Linting (prettier) and checking if the project builds is in my opinion part of testing. But feel free to suggest a different name though.

Nemikolh (approved):

.github/workflows/ci.yaml
d3lm: I mean we can keep it as is. From the top of my head I don't have a better name. Personally I would have different jobs and separate some of the stuff but like I said, we can keep it as is.

d3lm:
@SamVerschueren All status checks updated.

Nemikolh (approved):
Amazing work!! :muscle: :clap:

PR #61, opened 2024-06-10
feat: hot reload for files in webcontainer

Nemikolh:
This PR adds the ability to do an HMR-like update of files inside webcontainer when they are modified on disk without doing a full reload of the page. 

This should improve the DX around writing tutorial content a lot as modifying files that are expected to be in WebContainer is now a near-instant experience.

[hot reload of files in webcontainer demo](https://github.com/stackblitz/tutorialkit/assets/4008156/0fe9d333-42bf-453e-b2e6-b4111534b0f6)


packages/astro/src/webcontainer-files.ts
```diff
@@ -63,12 +64,12 @@ export class WebContainerFiles {
 
     await Promise.all(
       folders.map(async (folder) => {
-        const fileRef = getFilesRef(folder, { contentDir, templatesDir });
```
Nemikolh, line 66: I tried to normalize the name so it's now `filesRef` to refer to the string that is conceptually a "reference" to a set of files.

After doing that I also realised that the type `FilesRef` is incorrect and should probably be `FilesRefWithFileList` but I don't like that one so I haven't renamed it just yet.

cc @SamVerschueren

packages/runtime/src/store/index.ts
```diff
@@ -46,6 +47,68 @@ export class TutorialStore {
     this._previewsStore = new PreviewsStore(this._webcontainer);
     this._terminalStore = new TerminalStore(this._webcontainer, useAuth);
     this._runner = new TutorialRunner(this._webcontainer, this._terminalStore, this._stepController);
+
+    /**
+     * By having this code under `import.meta.hot`, it gets:
+     *  - Ignored on server side where it shouldn't run
+     *  - Discarded when doing a production build
```
d3lm, line 54:
```suggestion
     *  - ignored on server side where it shouldn't run
     *  - discarded when doing a production build
```

packages/types/src/files-ref.ts
```diff
@@ -0,0 +1,3 @@
+export function folderPathToFilesRef(pathToFolder: string): string {
+  return encodeURIComponent(pathToFolder.replaceAll('/', '-').replaceAll('\\', '-').replaceAll('_', '')) + '.json';
```
SamVerschueren, line 2: We could also use a regex if we want to instead of 2 `replaceAll()` calls. But there might be a reason you went with the string approach instead?

```suggestion
  return encodeURIComponent(pathToFolder.replaceAll(/[\/\\]+/g, '-').replaceAll('_', '')) + '.json';
```

Nemikolh: No good reason except that it was easier to write! Let's use the regex instead :+1:

packages/runtime/src/store/index.ts
```diff
@@ -46,6 +47,68 @@ export class TutorialStore {
...
+            case 'none': {
+              break;
+            }
+            case 'files': {
+              if (this._lesson?.files[0] === fileRef) {
+                shouldUpdate = true;
+
+                this._lesson.files[1] = Object.keys(result.files).sort();
+                this._lessonFiles = result.files;
+              }
```
SamVerschueren, line 73: I think we need to break here. Otherwise we're falling through to the `solution` case as well.

packages/runtime/src/lesson-files.ts
```diff
@@ -2,13 +2,41 @@ import type { Files, FilesRef, Lesson } from '@tutorialkit/types';
...
+
+    return {
+      type,
+      files,
+    };
+  }
+
   async getLessonTemplate(lesson: Lesson): Promise<Files> {
-    const templatePathname = `template-${lesson.data.template}`;
+    const templatePathname = `template-${lesson.data.template}.json`;
```
SamVerschueren, line 39: Was this a bug? 👀 ?

packages/runtime/src/lesson-files.ts
```diff
@@ -2,13 +2,41 @@ import type { Files, FilesRef, Lesson } from '@tutorialkit/types';
...
   private _map = new Map<string, Files>();
   private _templateLoadTask?: Task<Files>;
   private _templateLoaded: string | undefined;
 
+  async invalidate(fileRef: string): Promise<InvalidationResult> {
+    if (!this._map.has(fileRef)) {
+      return { type: 'none' };
+    }
+
+    const type = fileRef.startsWith('template-') ? 'template' : fileRef.endsWith('files.json') ? 'files' : 'solution';
```
SamVerschueren, line 22: Maybe create a util function to infer this type?

packages/astro/src/webcontainer-files.ts
SamVerschueren: I can't come up with something better than `FilesRefWithFileList`. So I would go for that name (for now). We could also just name it `FilesRefList` or `FilesRefMap`.

packages/runtime/src/store/index.ts
Nemikolh: Oh good catch!!

packages/runtime/src/lesson-files.ts
Nemikolh: It wasn't but I've normalized the logic around `filesRef`. In particular:
* `filesRef` is a string that semantically points to a list of files. It has to be fetched on the client to resolve to those files through the `LessonFilesFetcher`.

packages/runtime/src/lesson-files.ts
```diff
@@ -1,14 +1,42 @@
...
   private _map = new Map<string, Files>();
   private _templateLoadTask?: Task<Files>;
   private _templateLoaded: string | undefined;
 
+  async invalidate(filesRef: string): Promise<InvalidationResult> {
+    if (!this._map.has(filesRef)) {
+      return { type: 'none' };
+    }
+
+    const type = fromFilesRefToType(filesRef);
```
SamVerschueren, line 22: Super nitpick: Maybe rename the function to `getTypeFromFilesRef()`?

packages/runtime/src/lesson-files.ts
```diff
@@ -128,3 +156,7 @@ function convertToFiles(json: Record<string, string | { base64: string }>): File
 
   return result;
 }
+
+function fromFilesRefToType(filesRef: string): 'template' | 'files' | 'solution' {
+  return filesRef.startsWith('template-') ? 'template' : filesRef.endsWith('files.json') ? 'files' : 'solution';
```
SamVerschueren, line 161: Now that it's not inline anymore, I'd rather if a `if-else` statement here with early returns?

```ts
if (filesRef.startsWith('template-') {
  return 'template';
}

if (filesRef.endsWith('files.json') {
  return 'files';
}

return 'solution';
```

It's a bit longer, but much more readable (and maintainable) IMO.

packages/runtime/src/store/index.ts
Nemikolh: Whoopsie, I did!

SamVerschueren (approved):

PR #62, opened 2024-06-10
fix(logo): support multiple formats and remove styling requirements

Nemikolh:
This PR lets users specify a logo as either a JPEG, PNG or SVG file. It also fixes an issue where the SVG had to have an inlined `height: 100%` style otherwise it would end up being zero pixels in size.


packages/astro/src/default/components/Header.astro
```diff
@@ -1,15 +1,30 @@
...
+  const exists = fs.existsSync(path.join('public', logoFilename));
+
+  if (exists) {
+    logo = `/${logoFilename}`;
+    break;
+  }
+}
+
+if (!logo) {
+  console.warn(`No logo found in public/. Support filenames: ${logoFilenames.map((filename) => `'${filename}'`).join(', ')}`);
```
d3lm, line 27:
```suggestion
  console.warn(`No logo found in public/. Supported file names are: ${logoFilenames.map((filename) => `'${filename}'`).join(', ')}`);
```

d3lm: Instead of repeating the entire file name with every extension, what if we printed something like `logo.(svg|png|jpeg|jpg)` or something akin so it's a bit less repetitive? At least if the actual name is always the same.

Nemikolh: Oh very good point! Will update the PR with your suggestion 👍

SamVerschueren: Updated it.

SamVerschueren (approved):

packages/astro/src/default/components/Header.astro
d3lm: Nice 🔥

PR #63, opened 2024-06-11
fix: hydration error after runtime refactor

Nemikolh:
This PR fixes an error at runtime where React complains about an hydration mismatch between what was rendered on the server vs on the client.

This is because during SSR, we do not set the initial commands and thus the `steps` are empty while on the client they are set sooner. I wouldn't have expected that to be the case but the easier here is to just split `runCommands` in two functions:
* `setCommands` which is similar to `setPreviews` and `setTerminalConfiguration`.
* `runCommands` which interact with WebContainer

The error fixed by this PR:

image

### DIFF
--- a/packages/runtime/src/tutorial-runner.ts
+++ b/packages/runtime/src/tutorial-runner.ts
@@ -77,7 +77,7 @@ export class TutorialRunner {
    *
    * This function is safe to call server side.
    *
-   * To actually run them in webcontainer see `runCommands`.
+   * To actually run them in WebContainer see `runCommands`.
    *
    * @param commands The commands schema.
    */
--- a/packages/runtime/src/tutorial-runner.ts
+++ b/packages/runtime/src/tutorial-runner.ts
@@ -74,7 +74,7 @@ export class TutorialRunner {
 
   /**
    *
-   * @param commands Commands schema
+   * @param commands The commands schema.
    */
   setCommands(commands: CommandsSchema) {
     const newCommands = new Commands(commands);
--- a/packages/runtime/src/tutorial-runner.ts
+++ b/packages/runtime/src/tutorial-runner.ts
@@ -38,7 +38,7 @@ interface LoadFilesOptions {
   signal?: AbortSignal;
 }
 
-interface RunCommandsOptions extends CommandsSchema {
+interface RunCommandsOptions {
   /**
    * Abort the previous run commands operation.
    *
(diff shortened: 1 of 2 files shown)


SamVerschueren (approved):

packages/runtime/src/tutorial-runner.ts
```diff
@@ -71,6 +72,22 @@ export class TutorialRunner {
     private _stepController: StepsController,
   ) {}
 
+  /**
+   *
```
d3lm, line 76: Can we add a description here as well?

packages/runtime/src/tutorial-runner.ts
```diff
@@ -71,6 +72,22 @@ export class TutorialRunner {
...
+   * @param commands Commands schema
```
d3lm, line 77:
```suggestion
   * @param commands The commands schema.
```

packages/runtime/src/tutorial-runner.ts
```diff
@@ -71,6 +72,22 @@ export class TutorialRunner {
...
+   */
+  setCommands(commands: CommandsSchema) {
+    const newCommands = new Commands(commands);
+    let anyChange = this._changeDetection(commands);
```
d3lm, line 81: Newline to separate const from let?

Nemikolh: Oh actually `let` wasn't needed! I changed it to a `const`.

packages/runtime/src/tutorial-runner.ts
```diff
@@ -71,6 +72,27 @@ export class TutorialRunner {
     private _stepController: StepsController,
   ) {}
 
+  /**
+   * Set the commands to run. This updates the reported `steps` if any have changed.
+   *
+   * This function is safe to call server side.
+   *
+   * To actually run them in webcontainer see `runCommands`.
```
d3lm, line 80:
```suggestion
   * To actually run them in WebContainer see `runCommands`.
```

d3lm (commented):
Tiny tiny nitpick 🙉 Otherwise LGTM 👍

packages/runtime/src/tutorial-runner.ts
Nemikolh: Ah good catch! :pray:

d3lm (approved):

PR #64, opened 2024-06-11
feat: allow custom components that modify the tutorial state

Nemikolh:
This PR makes it possible to create custom components that can interact with the state of the tutorial when using the default Astro experience that uses React.

Here's an example:

* You add `src/components/ResetButton.tsx`

```tsx

export default function ResetButton() {
  return (
    <button
      className="bg-red-5 hover:bg-red-6 rounded-lg px-2 py-1 text-sm text-white"
      onClick={() => tutorialStore.reset()}
    >
      Reset
    </button>
  );
}
```

* You use it in your lesson `content.mdx` like this:

```mdx
---
type: lesson
title: My lesson
---
import ResetButton from '@components/ResetButton';

<ResetButton client:load />
```

The `tutorialkit:store` module exposes `tutorialStore` and the `tutorialkit:core` module exposes `webcontainer` which lets you modify the state of the tutorial and access the low level WebContainer API respectively. The first one should be preferred but the latter gives you full control over the WebContainer instance if you need it.

On the `tutorialStore` instance, you have access to most of the actionable state. You also can perform a few useful actions such as:

 * `updateFile`: to update the content of a file.
 * `setSelectedFile`: to switch the selected file in the file tree
 * `terminalConfig.panels[<terminal-index>].write('vitest')`: to write content in a writeable terminal

This will likely evolve over time, we'll probably provide better APIs. If this looks exciting to you, let us know what you're looking to do! :smiley:


d3lm (approved):

packages/astro/src/vite-plugins/state.ts
```diff
@@ -0,0 +1,39 @@
...
+  resolveId(id) {
+    if (id === virtualModuleId) {
+      return resolvedVirtualModuleId;
+    }
+  },
+  async load(id) {
+    if (id === resolvedVirtualModuleId) {
+      const pathToInit = path.join(__dirname, 'default', 'components', 'webcontainer.ts');
+
+      return `
```
SamVerschueren, line 34: So if nothing is returned, I assume it just goes to the next loader?

Nemikolh: Yup! See https://vitejs.dev/guide/api-plugin.html#virtual-modules-convention for examples from the docs.

packages/astro/src/vite-plugins/state.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * A plugin that let users write components that can interact with the tutorial
+ * state, indirectly acting on the editor and WebContainer.
+ *
+ * The virtual module can be imported as:
+ *
+ * ```ts
+ * import { tutorialStore } from 'tutorialkit:state';
```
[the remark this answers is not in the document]
Nemikolh: I feel like `state` express better what you get like `astro:content`. I personally find `core` a bit too generic.

packages/astro/src/vite-plugins/core.ts
```diff
@@ -0,0 +1,46 @@
...
+export const tutorialkitCore: VitePlugin = {
+  name: 'tutorialkit-core-virtual-mod-plugin',
+  resolveId(id) {
+    if (id === virtualModuleId) {
+      return resolvedVirtualModuleId;
+    }
+  },
+  async load(id) {
+    if (id === resolvedVirtualModuleId) {
+      const pathToInit = path.join(__dirname, 'default', 'components', 'webcontainer.ts');
```
SamVerschueren, line 39: I would just go with

```suggestion
      const pathToInit = path.join(__dirname, 'default/components/webcontainer.ts');
```

packages/astro/src/vite-plugins/store.ts
```diff
@@ -0,0 +1,40 @@
...
+export const tutorialkitStore: VitePlugin = {
+  name: 'tutorialkit-store-virtual-mod-plugin',
+  resolveId(id) {
+    if (id === virtualModuleId) {
+      return resolvedVirtualModuleId;
+    }
+  },
+  async load(id) {
+    if (id === resolvedVirtualModuleId) {
+      const pathToInit = path.join(__dirname, 'default', 'components', 'webcontainer.ts');
```
SamVerschueren, line 32: Same as above.

```suggestion
      const pathToInit = path.join(__dirname, 'default/components/webcontainer.ts');
```

packages/template/src/content/tutorial/1-basics/1-introduction/2-foo/content.mdx
```diff
@@ -11,6 +11,10 @@ terminal:
   panels: 'terminal'
 ---
 
+import ExampleComponent from '@components/ExampleComponent';
+
+<ExampleComponent client:load />
```
SamVerschueren, line 16: I think we want to remove this?

SamVerschueren (approved):
Small things, but generally looks good!

SamVerschueren (approved):

PR #65, opened 2024-06-12
feat: add lezer support for wast via codemirror wast package

EmNudge:
Adds support for the wat language via [@codemirror/lang-wast](https://github.com/codemirror/lang-wast) ([context](https://x.com/elmd_/status/1800053726075310249))

We may want to explore exposing support for users to add their own lezer grammars and language servers. In the case of, perhaps a custom language, we might not want to merge it into the package directly but still want to allow support without monkey-patching the package.

### DIFF
--- a/.changeset/angry-man-yells.md
+++ b/.changeset/angry-man-yells.md
@@ -1,0 +1,5 @@
+---
+"@tutorialkit/astro": patch
+---
+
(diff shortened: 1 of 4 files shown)


SamVerschueren:
Thanks @EmNudge for this PR! Love to see it ❤️ !

SamVerschueren (approved):

PR #67, opened 2024-06-12
fix: support inheritance for `editor`/`focus` and fix bug with logo

Nemikolh:
This PR makes sure that inheritance is supported for both `editor` and `focus`. It also fix a bug with the logo where it tried to fetch `svg` instead of the logo filename.


packages/astro/src/default/components/Header.astro
```diff
@@ -14,8 +14,9 @@ const LOGO_EXTENSIONS = [
 
 let logo;
 
-for (const logoFilename of LOGO_EXTENSIONS) {
-  const exists = fs.existsSync(path.join('public', `logo.${logoFilename}`));
+for (const logoExt of LOGO_EXTENSIONS) {
+  const logoFilename = `logo.${logoExt}`;
+  const exists = fs.existsSync(path.join('public', logoFilename));
```
SamVerschueren, line 19: Woops, that's my fault 🤦 . I ran the project and the only thing I looked at was the warning...

SamVerschueren (approved):

packages/astro/src/default/components/Header.astro
Nemikolh: No worries! :raised_hands:

PR #68, opened 2024-06-12
feat: make the logo link configurable

Nemikolh:
This PR makes the logo link configurable so that it can point anywhere.

To use this feature add the following property to your `tutorial/meta.md` file:

```md
---
type: tutorial
logoLink: https://example.com
---
```

> This PR also default the collection schema to `strict()` so that unknown properties are caught.


packages/types/src/schemas/tutorial.ts
```diff
@@ -3,6 +3,7 @@ import { webcontainerSchema } from './common.js';
 
 export const tutorialSchema = webcontainerSchema.extend({
   type: z.literal('tutorial'),
+  logoLink: z.string().optional(),
```
SamVerschueren, line 6: Not sure about the `logoLink` name. Should we just call it `homepage` or `home` instead? I'm fine with keeping it though. You know, naming...

packages/astro/src/default/components/Header.astro
```diff
@@ -5,6 +5,12 @@ import { LoginButton } from './LoginButton';
 import { ThemeSwitch } from './ThemeSwitch';
 import { useAuth } from './setup';
 
+interface Props {
+  logoLink: string;
```
SamVerschueren, line 9: Given it's optional in the schema, we should probably also make it optional here? And not render a link at all of the option is not set.

SamVerschueren: Or default to `href="/"`?

SamVerschueren (approved):
Awesome work @Nemikolh 🙏 .

packages/types/src/schemas/tutorial.ts
Nemikolh: Yeah I found `logoLink` to be more explicit about what that property does but it doesn't look great.

I don't know what to think about `homepage` or `home` because you can have it point to anything like `https://buymeacoffee.com/...`.

packages/astro/src/default/components/Header.astro
Nemikolh: It's defaulted to `/` if not set in [`[...slug].astro`](https://github.com/stackblitz/tutorialkit/pull/68/files#diff-9402baf5c6f38adc001724a8d4876230199fa0ff082bb036e839a60ad3195f65R29)

d3lm (commented):
Love it that the logo link is configurable. One thing I was wondering if we should not use the meta file inside the content folder and instead have various tutorial wide things which are less related to the actual tutorial content be configured through a config file like `tutorialkit.config.ts` which exports a POJO. That would also be more flexible and in terms of DX it's a little more convenient to work with than yaml. Also thinking about the top bar config etc. and maybe folks wanted to do some dynamic stuff and read data from other files? Just thinking out loud.

Nemikolh:
Very good point!

On the note of having a `tutorialkit.config.ts`, what do you think about having the options be provided directly to the astro integrations, (`tutorialkit({ ... })`) instead ?

This way end users can decide to have a separate file or not and it's one less file for end users.

SamVerschueren:
The reason why I would be more in favour of a `tutorialkit.config.ts` is because it's then separated from Astro and the underlying tech. I mean, we could support both though 🤷 . Also given we hide the `astro.config.ts` by default... It would also make configuring the github importer easier because you don't even have an `astro.config.ts`.

Nemikolh:
> The reason why I would be more in favour of a `tutorialkit.config.ts` is because it's then separated from Astro and the underlying tech.

AFAIK, the default TutorialKit experience is only ever going to support Astro and I don't see us supporting alternative techs.

If we embraced it entirely, then we can also benefit from the full Astro ecosystem where users can bring any Astro integrations that they want.

Lastly, we want to keep the Astro config around for the Astro VSCode extension. So this is a file that we can't really remove.

> It would also make configuring the github importer easier because you don't even have an `astro.config.ts`.

I've been changing my mind on that one recently, I don't think anymore that we should assume the branch does not have an `astro.config.ts` anymore. The main reasons being: 
 * To let users bring their own integrations
 * It's not a security issue, the whole thing is: components are executed server side. And if we have a `tutorialkit.config.ts` then it's the same anyway.
 * It one less build mode to support

Curious to hear your thoughts on how you envision the github importer.

packages/template/public/logo-dark.svg
Nemikolh, line 1: As per our discussion with @SamVerschueren, instead of relying on `currentColor` we have the color as part of the logo (as this should often be the case for logos) and we lookup for a `logo-dark` variant that if not found we fallback to `logo`.

packages/astro/src/default/components/Logo.astro
```diff
@@ -0,0 +1,43 @@
...
+  return logo;
+}
+
+const { logoLink } = Astro.props;
+
+const logo = readLogoFile('logo');
+const logoDark = readLogoFile('logo-dark') ?? logo;
+
+if (!logo) {
+  console.warn(`No logo found in public/. Supported filenames are: logo.(${LOGO_EXTENSIONS.join('|')})`);
```
d3lm, line 33: Should we also mention that we look for a `logo-dark`?

packages/types/src/schemas/tutorial.ts
d3lm, line 6: Did we now decide whether to use the TK integration for configuration or a separate file? Personally I'd prefer if we decided that and maybe added support for that in this PR before we go back and forth on frontmatter configurations? I mean it's always a little cumbersome for users if they have to update and if we have folks that use the frontmatter and we switch it to a config file very soon then updating isn't as smooth even if it's a small change but if we can avoid it that'd be great.

Nemikolh: We haven't decided on anything yet so I haven't changed it.

My opinion? I think that property should be in a frontmatter rather than in the configuration of the TK integration or a separate configuration file.

Here's why I think so:

 * Having it in the frontmatter instead of a configuration file means that there's no need for the vite server to be restarted. Which makes sense given I'm only setting an href on an element in the UI.
 * In the last sync meeting, we've been discussing of having a `TopBar.astro` (or equivalent) that could be picked up automatically to give full control over how the top bar is configured. This would not require any configuration change, so having one for the logoLink is weird
 * We have `editor: false` which is also a frontmatter setting right now and also a UI concern so I see arguments of having `logoLink` in a config to also apply to `editor`.

Lastly, I would keep configuration at the root of the repo to be focused on stuff like port, headers, routes, authentication. Today we have:
* `enterprise` to point at a StackBlitz Enterprise instance
* `defaultRoutes` which when set to false let Astro lookup routes in `src/pages` useful when you want full customizablility.

If possible I'd like to keep it this way. YMMV.

packages/astro/src/default/components/Logo.astro
Nemikolh: I was thinking of leaving it to the docs given that it's optional. Or maybe we could say something like:

```js
`
No logo found in public/. 
Supported filenames are: logo.(${LOGO_EXTENSIONS.join('|')})
(A logo-dark.(${LOGO_EXTENSIONS.join('|')}) will be used if found for the dark theme)`
```

Wdyt?

packages/astro/src/default/components/Logo.astro
```diff
@@ -0,0 +1,43 @@
...
+}
+---
+
+<a
+  href={logoLink}
+  class="flex items-center text-tk-elements-topBar-logo-color hover:text-tk-elements-topBar-logo-colorHover"
+>
+  {logo && <img class="h-5 w-auto dark:hidden" src={logo} />}
```
SamVerschueren, line 41: Should `logo` also fallback to `logo-dark`? I'm just imagining if someone removed `logo.svg` but still has `logo-dark.svg`? Although, maybe that doesn't make sense because it's explicitly `logo-dark`...

SamVerschueren (approved):

packages/astro/src/default/components/Logo.astro
Nemikolh: I've added a mention for `logo-dark`. :+1:

packages/astro/src/default/components/Logo.astro
Nemikolh: We discussed this and agreed to keep the current logic where we don't default to `logo-dark` if no `logo` is present.

PR #70, opened 2024-06-12
chore: bump version and add CHANGELOG.md

Nemikolh:
This PR introduces `conventional-changelog` to update automatically the `CHANGELOG.md` file based on the git history.

To create a new version in the changelog, one must do:
 1. Update the `version` field in every package
 2. Run `pnpm changelog`
 3. Check the update to `CHANGELOG.md` and add shoutouts for our awesome external contributors :partying_face: 
 4. Open a PR
 5. When the PR is merged and the package is published, update a new tag with:

```bash
git tag -a $VERSION $COMMIT_HASH -m "TutorialKit $VERSION"
git push --tags
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,11 @@ Special thanks to @EmNudge and @morinokami for their first contributions!! 🥳
 ### Bug Fixes
 
 * feature request issue template ([cc8423a](https://github.com/stackblitz/tutorialkit/commit/cc8423abe2c613f82d73179ca82f39e4ac0929c9))
+* generate root changelog + changelog per package ([24d4131](https://github.com/stackblitz/tutorialkit/commit/24d4131ff5ffca9fde614cb3dd7682d6eca60433))
 * hydration error after runtime refactor ([#63](https://github.com/stackblitz/tutorialkit/issues/63)) ([8f90338](https://github.com/stackblitz/tutorialkit/commit/8f9033816cd122be49ade2b85e0040469ed9fb1c))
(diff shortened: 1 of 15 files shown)


d3lm (commented):
Generally I like this but this generates only a single changelog (right?) and it'd be nice to have a changelog for all packages that we publish separately IMO. From this changelog it's not clear which feature or bug belongs to which package (unless I click on something I check myself). That's what I mentioned on Slack where a single changelog could work if we enforced scopes where the scope is the package so it's immediately clear where it belongs to.

Nemikolh:
@d3lm Ah got it! Makes tons of sense :pray: 

The `conventional-changelog` tool supports providing a `--commit-path`. With this flag I think we could run it once per package and have it update the changelog for only that package.

Do you think that's worth pursuing?

d3lm:
Yea I think that would be cool! Having one changelog for every package we publish is prolly desirable and if we can use `--commit-path` to do that and run it once per every package that would be cool! Is it possible to include the type of the commit as well? Like `feat` etc. cause I do like that a lot in the Node.js changelog for instance cause I can quickly ignore some type of commits or look for commits of a specific type which I have done plenty of times for upgrading WebContainer.

Nemikolh:
> Is it possible to include the type of the commit as well?

From my observation, with the `angular` preset it seems that the tool uses the type to decide where the commit should be positioned in the `CHANGELOG.md`. So for instance, for `feat`, the commit ends up in the `## Features` section:

```md
## Features

<!-- A commit like "feat: foobar (#123)" seems to land here: -->
* foobar ([#123](<link PR>)) ([<commit hash>](<link commit>))
```

They seems to have support for other preset and the ability to have our own! :exploding_head: 

I can experiement with the presets and let you know what we get for each if you want!

d3lm:
I think if it's separated into separate sections then it should be fine 👍 Thanks for checking.

SamVerschueren (approved):
Looks great!

PR #73, opened 2024-06-14
test: fix tests

SamVerschueren:
I'm still trying to figure out why on earth the compilation suddenly started failing. The reason for it is just ridiculous.

I found out that every module we load in `languages.ts` and which we call with an object, cannot use the shorthand return syntax.

**wrong**
```ts
import('@codemirror/lang-javascript').then((module) => module.javascript({ typescript: true }));
```

**correct**
```ts
import('@codemirror/lang-javascript').then((module) => {
  return module.javascript({ typescript: true });
});
```

I'll buy anyone a drink that can explain me why on earth the compiler failed on that syntax.

### DIFF
--- a/packages/components/react/src/CodeMirrorEditor/languages.ts
+++ b/packages/components/react/src/CodeMirrorEditor/languages.ts
@@ -5,7 +5,9 @@ export const supportedLanguages = [
     name: 'TS',
     extensions: ['ts'],
     async load() {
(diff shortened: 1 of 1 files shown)


SamVerschueren:
It was a bug in Vite and was fixed here https://github.com/vitejs/vite/pull/17479.

PR #74, opened 2024-06-14
feat(store): fix current document and add onDocumentChanged

SamVerschueren:
This PR fixes an issue where `store.updateFile()` didn't reflect the new content in the editor.

It also adds a new method to the store called `onDocumentChanged()` which you can use to subscribe to changed to a document. This allows you to do clever things in a custom component like listening to the changes of a `css` file and dynamically injecting that into the page.

### DIFF
--- a/packages/runtime/src/store/editor.ts
+++ b/packages/runtime/src/store/editor.ts
@@ -1,5 +1,5 @@
 import type { FilesRefList, Files } from '@tutorialkit/types';
-import { atom, computed } from 'nanostores';
+import { atom, map, computed } from 'nanostores';
 
 export interface EditorDocument {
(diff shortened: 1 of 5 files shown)


packages/runtime/src/store/editor.ts
```diff
@@ -90,9 +93,46 @@ export class EditorStore {
...
+          callback(document);
+
+          unsubscribeFromDocuments();
+        });
+      }
+    });
+
+    return () => {
+      unsubscribeFromDocuments();
+
```
d3lm, line 134: Nit: Remove newline?

d3lm (approved):

Nemikolh (approved):
Nice :tada:

PR #75, opened 2024-06-17
chore: add tests to runtime

Nemikolh:
This PR adds tests to the `@tutorialkit/runtime` package and add an helper package, `webcontainer-api-mock` to help testing code that uses the WebContainer API. 

It's minimal at this point but the idea is that it will grow to support `@tutorialkit/components-react` over time.

I'm also planning to add more tests for the `runtime` but I wanted to open the PR to start getting feedback on the approach taken with the `webcontainer-api-mock` package.


packages/runtime/tests/webcontainer/terminal-config.spec.ts
SamVerschueren, line 1: I think I like it more if the spec file is next to the implementation file. So by moving this file to `packages/runtime/webcontainer/terminal-config.spec.ts`. Same for the other files.

The only reason I made a separate package in `cli` is because it's testing the entire thing.

packages/webcontainer-api-mock/tsconfig.json
SamVerschueren, line 1: If we move the files, is this still needed then?

SamVerschueren (commented):
Some small comments, but looks very good!

packages/webcontainer-api-mock/package.json
```diff
@@ -0,0 +1,23 @@
...
+  ],
+  "scripts": {
+    "build": "tsc -b"
+  },
+  "devDependencies": {
+    "@webcontainer/api": "1.2.0",
+    "typescript": "^5.4.5",
+    "vitest": "^1.6.0"
+  }
+}
```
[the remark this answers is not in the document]
Nemikolh: I've been thinking that this might be useful to others that write tests for code that uses the WebContainer API and so maybe we should publish it as a result?

Or do you think it's best to have it named `test-utils` and postpone to later the idea of publishing it?

packages/runtime/tests/webcontainer/terminal-config.spec.ts
Nemikolh: I actually like what you said because it's what you do in Rust, I wasn't sure if we should follow that pattern here too. I'll move them all :+1:

packages/webcontainer-api-mock/tsconfig.json
Nemikolh: Why wouldn't it be? Or did you mean to talk about the `tests/tsconfig.json` file?

packages/webcontainer-api-mock/package.json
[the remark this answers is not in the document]
Nemikolh: Good point :+1: Let's go with `test-utils` then.

SamVerschueren (commented):
Just a small comment, other than that, really love the fact that we have much more tests!

.prettierignore
```diff
@@ -1,3 +1,4 @@
 pnpm-lock.yaml
+.astro
```
SamVerschueren, line 2: Shouldn't `.astro` files be gitignored?

packages/test-utils/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "test-utils",
+  "version": "1.0.0",
+  "description": "Test utilities for TutorialKit",
+  "type": "module",
+  "private": true,
+  "bugs": "https://github.com/stackblitz/tutorialkit/issues",
+  "homepage": "https://github.com/stackblitz/tutorialkit",
+  "license": "MIT",
+  "exports": {
```
SamVerschueren, line 10:
```suggestion
  "exports": {
```

SamVerschueren (commented):
Very nice! Just two very small questions/remarks.

.prettierignore
Nemikolh: The [`.astro` folder](https://github.com/stackblitz/tutorialkit/blob/db233c7b37a787677d205586a88663913e71e99f/packages/template/.gitignore#L2) is ignored already, I noticed that prettier was giving a warning locally because it doesn't ignore it.

It doesn't produce that warning on CI because we run prettier before running the build script. If the order was reversed, the CI would fail which would be a bit silly.

SamVerschueren: OOh weird. Ok, then it's fine.

SamVerschueren (approved):
Awesome work @Nemikolh 🙏 !

PR #76, opened 2024-06-18
feat(terminal): add support for redirects and allow specific commands

SamVerschueren:
This PR adds support for setting `allowRedirects` and `allowCommands` when defining the terminal panels. You can define the options for all terminals in the lesson, or separately per terminal.

I also removed defining the panels with a tuple where the second argument is a config object, because the syntax was bad... This means that you can't do this anymore

```yaml
terminal:
  - [terminal, { name: 'Terminal Name' }]
```

But instead do

```yaml
terminal:
  - type: terminal
    name: 'Terminal Name'
```

or providing the name as the second argument is still supported.

```yaml
terminal:
  - [terminal, 'Terminal Name']
```

With the new options, it looks like this

```yaml
terminal:
  panels:
    - terminal
    - output
  allowRedirects: true
  allowCommands:
    - ls
    - echo
```

```yaml
terminal:
  panels:
    - type: terminal
      allowRedirects: true
      allowCommands:
        - ls
    - type: terminal
      allowRedirects: false
      allowCommands:
        - echo
    - output
```


Nemikolh:
Oh could we keep the `[terminal, 'Terminal Name']` shorthand for only specifying the type and the title? I quite like that syntax and it's also supported by the `previews` field and it's nice to support it for both.

Note that this syntax is optional, we don't force anyone to use it but for those that enjoys it (like me) it's pretty cool to support it.

SamVerschueren:
I'm not a big fan of having 100 different syntaxes for the same thing. But I can add it back if we want to keep it though. Wasn't aware it was supported by the previews as well, because in that case it kinda makes sense I guess.

SamVerschueren:
Added it back, but removed the `[type, object]` syntax though. So the second argument can only be the name, not the config object anymore.

Nemikolh (approved):
This is really nice, love it :heart:

packages/runtime/src/webcontainer/terminal-config.ts
```diff
@@ -59,9 +66,10 @@ export class TerminalPanel implements ITerminal {
 
   constructor(
     readonly type: TerminalPanelType,
-    name?: string,
-    id?: string,
+    private readonly options?: TerminalPanelOptions,
   ) {
+    let name = options?.name;
```
Nemikolh, line 71: Wondering if we should have named this `title` like for `previews`.

packages/runtime/src/webcontainer/terminal-config.ts
```diff
@@ -196,19 +211,24 @@ function normalizeTerminalConfig(config?: TerminalSchema): NormalizedTerminalCon
...
+            name: panel[1],
+          });
         } else {
-          name = config.name;
-          id = config.id;
+          terminalPanel = new TerminalPanel(panel.type, {
+            id: panel.id,
+            name: panel.name,
+            allowRedirects: panel.allowRedirects ?? config.allowRedirects,
+            allowCommands: panel.allowCommands ?? config.allowCommands,
```
Nemikolh, line 227: Oh this is pretty cool that we can define it once for all terminals :pinched_fingers:

Nemikolh (approved):
Really nice!! Good work on this :clap:

packages/runtime/src/webcontainer/terminal-config.ts
SamVerschueren: Oh yeah, I agree. Let's change it?

docs/tutorialkit.dev/src/content/docs/guides/configuration.mdx
```diff
@@ -38,16 +38,16 @@ Defines which file should be opened in the [code editor](/guides/ui/#code-editor
 <PropertyTable inherited type="string" />
 
 ##### `previews`
-Configure which ports should be used for the previews allowing you to align the behavior with your demo application's dev server setup. If not specified, the lowest port will be used. 
+Configure which ports should be used for the previews allowing you to align the behavior with your demo application's dev server setup. If not specified, the lowest port will be used.
 
-You can optionally provide these as an array of tuples where the first element is the port number and the second is the name of the preview, or as an object.
+You can optionally provide these as an array of tuples where the first element is the port number and the second is the title of the preview, or as an object.
```
SamVerschueren, line 43: Renamed `name` to `title` in this sentence.

docs/tutorialkit.dev/src/content/docs/guides/configuration.mdx
```diff
@@ -91,12 +95,14 @@ The `Terminal` type has the following shape:
...
+    activePanel?: number,
+    allowRedirects?: boolean,
+    allowCommands?: string[]
 }
 
 type TerminalPanel = TerminalType
-    | [type: TerminalType, name: string]
-    | [type: TerminalType, { name: string, id: string }]
+    | [type: TerminalType, title: string]
+    | { type: TerminalType, title?: string, id?: string, allowRedirects?: boolean, allowCommands?: boolean }
```
SamVerschueren, line 105: It's now `title` instead of `name`.

packages/types/src/schemas/common.ts
```diff
@@ -101,7 +101,11 @@ export const terminalSchema = z.union([
             let output = 0;
 
             for (const value of arg) {
-              if (value === 'output' || (Array.isArray(value) && value[0] === 'output')) {
+              if (
+                value === 'output' ||
+                (Array.isArray(value) && value[0] === 'output') ||
+                (typeof value === 'object' && (value as any).type === 'output')
```
Nemikolh, line 107: Oh good catch!

Nemikolh (approved):
Looks really good!! :muscle: Really good work on this! :tada:

PR #77, opened 2024-06-19
refactor(cli): only install and start project after all questions

SamVerschueren:
This PR refactors the CLI (a bit).

- It will now always ask which package manager to use.
- The last question has now changed to `Install dependencies and start project?`
- Added a new `--start` flag so you could do `--install --no-start` for instance.

I'll add some additional comments in the code.


packages/cli/src/commands/create/git.ts
```diff
@@ -6,12 +6,12 @@ import { pkg } from '../../pkg.js';
 import { warnLabel } from '../../utils/messages.js';
 import { runShellCommand } from '../../utils/shell.js';
 import { assertNotCanceled, runTask } from '../../utils/tasks.js';
-import { DEFAULT_VALUES, type CreateOptions } from './options.js';
+import { DEFAULT_VALUES, readFlag, type CreateOptions } from './options.js';
 
 export async function initGitRepo(cwd: string, flags: CreateOptions) {
-  let shouldInitGitRepo = flags.git ?? DEFAULT_VALUES.git;
+  let shouldInitGitRepo = readFlag(flags, 'git');
```
SamVerschueren, line 12: I created a small `readFlag()` utility which returns the value of the flag, or if the flag doesn't have a value, it will return the default ONLY if `flags.default` is `true`. It makes the logic a bit easier because you know that if the value is `undefined`, you can prompt the user for the question because it also means `--defaults` was not provided.

packages/cli/src/commands/create/index.ts
```diff
@@ -46,6 +49,16 @@ export async function createTutorial(flags: yargs.Arguments) {
     return 0;
   }
 
+  applyAliases(flags);
+
+  try {
+    verifyFlags(flags);
```
SamVerschueren, line 55: It's not possible to do `--start --no-install`. You can't start the project without installing the dependencies. If that happens, we throw. This method could have additional checks in the future.

packages/cli/src/commands/create/package-manager.ts
SamVerschueren, line 1: The logic here is extracted from `dependencies.ts`.

packages/cli/src/commands/create/install-start.ts
```diff
@@ -0,0 +1,44 @@
+import * as prompts from '@clack/prompts';
+import { DEFAULT_VALUES, readFlag, type CreateOptions } from './options.js';
+import { assertNotCanceled } from '../../utils/tasks.js';
+
+export async function installAndStart(flags: CreateOptions) {
```
SamVerschueren, line 5: This function combines both the `--install` and `--start` flags.

packages/cli/src/commands/create/install-start.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  if (installDeps) {
+    if (startProject === false) {
+      // the user wants to install the dependencies but expicitly not start the project
+      return { install: true, start: false };
+    } else {
+      // the user explicitly wants to install the dependencies, but we don't know if they want to start the project
+      const answer = await prompts.confirm({
+        message: 'Start project?',
```
SamVerschueren, line 26: If `--install` is provided, but not `--start`, we change the question to just `Start project?`.

packages/cli/src/commands/create/index.ts
```diff
@@ -131,21 +142,55 @@ async function _createTutorial(flags: CreateOptions) {
...
+    let message = 'Please wait while we install the dependencies and start your project...';
 
-  prompts.outro(`You're all set!`);
+    if (install && !start) {
+      // change the message if we're only installing dependencies
+      message = 'Please wait while we install the dependencies...';
 
-  console.log('Until next time 👋');
+      // print the next steps without the install step in case we only install dependencies
+      printNextSteps(dest, selectedPackageManager, true);
```
SamVerschueren, line 165: If we install the dependencies, but we don't start the project, we still print the next steps without the `install` step.

```
◆  Tutorial successfully created!
│
│  Next Steps
│
◇  1. cd ./dry-king - Navigate to project
│
◇  2. npm run dev - Start development server
│
◇  3. Head over to http://localhost:4321
│
└  Please wait while we install the dependencies...
```

packages/cli/src/commands/create/index.ts
```diff
@@ -131,21 +142,55 @@ async function _createTutorial(flags: CreateOptions) {
...
+  }
+}
+
+async function startProject(cwd: string, packageManager: PackageManager, flags: CreateOptions, startProject: boolean) {
+  if (flags.dryRun) {
+    const message = startProject
+      ? 'Skipped dependency installation and project start'
+      : 'Skipped dependency installation';
+
+    console.warn(`${warnLabel('DRY RUN')} ${message}`);
```
SamVerschueren, line 186: We don't do anything in case of a dry run.

```
◆  Tutorial successfully created!
│
└  Please wait while we install the dependencies and start your project...

 DRY RUN  Skipped dependency installation and project start
```

Nemikolh (approved):
Looks good :tada:

PR #78, opened 2024-06-19
docs: add tutorialkit.dev website

SamVerschueren:
This PR moves the website from `tutorialkit.dev` into this repository.

I moved it inside the `docs/tutorialkit.dev` folder. The reason I didn't move it directly into `docs` is because we also want to move the demo inside this repository, which I would put under `docs/demo`. I'm open to suggestions though. We could also do `docs` and `demo` in the root of the project for instance.


docs/tutorialkit.dev/.gitignore
Nemikolh, line 1: This file can probably be removed?

package.json
```diff
@@ -7,6 +7,8 @@
     "prepare": "is-ci || husky install",
     "template:dev": "TUTORIALKIT_DEV=true pnpm run build && pnpm run --filter=tutorialkit-starter dev",
     "template:build": "pnpm run build && pnpm run --filter=tutorialkit-starter build",
+    "tkit": "pnpm run --filter=tutorialkit.dev dev",
```
Nemikolh, line 10: What if this was `"docs"` instead of `"tkit"`?

Nemikolh (approved):
This looks great! :tada:

package.json
SamVerschueren: So first I had `docs`. It had 2 minor "issues"

1. You can't run `pnpm docs`, because it opens the pnpm documentation. So you need to run `pnpm run docs`.
2. If we are moving the demo over to this repo as well, it might be confusing?

I don't know, it's not big of a deal and we could always just have `pnpm run demo` and `pnpm run docs`. Maybe that's better?

Nemikolh: Oh I quite like `pnpm run demo` and `pnpm run docs`! :star2: 

It's a bummer that `pnpm docs` doesn't work. Maybe we could have `docs-site` instead? 

Although I think I prefer `pnpm run docs`.

docs/tutorialkit.dev/.gitignore
SamVerschueren: I think I'll keep it for `tutorialkit.dev` specific files/folders. Like this one also has `.env` and `.env.production`. I don't think they are used though, but maybe they are?

The `.astro` folder is something specific to this folder and I don't think it makes much sense to move that to the root?

So the file would then look like this

```
# generated types
.astro

# environment variables
.env
.env.production
```

package.json
SamVerschueren: Yeah I agree. I'll update it.

docs/tutorialkit.dev/.gitignore
Nemikolh: Ah good point! Yep let's simplify it but keep it :+1:

Nemikolh (approved):
:star_struck:

PR #79, opened 2024-06-19
feat: allow ordering to be config based in addition to folder based

Nemikolh:
This PR makes it possible to specify the ordering in the metadata of the `tutorial`, `part` or `chapter`. If it isn't specified it fallback to the current behaviour and when it's used it prints warning if it finds any issues:

image

## How to specify the order

In `tutorial/meta.md`:

```yaml
type: tutorial
parts:
  - foo         # Must match the name of the folder containing the part
  - bar
```

In `tutorial/foo/meta.md`:

```yaml
type: part
chapters:
  - chapterfoo  # Must match the name of the folder containing the chapter
```

In `tutorial/foo/chapterfoo/meta.md`:

```yaml
type: chapter
lessons:
  - welcome    # Must match the name of the folder containing the lesson
```


packages/astro/src/default/pages/index.astro
```diff
@@ -3,10 +3,9 @@ import { getTutorial } from '../utils/content';
 
 const tutorial = await getTutorial();
 
-const parts = Object.values(tutorial.parts);
-const part = parts[0];
-const chapter = part.chapters[1];
-const lesson = chapter.lessons[1];
```
SamVerschueren, line 9: Why was this 1-based?

packages/astro/src/default/pages/index.astro
```diff
@@ -3,10 +3,9 @@ import { getTutorial } from '../utils/content';
...
+const part = tutorial.parts[tutorial.firstPartId!];
+const chapter = part.chapters[part?.firstChapterId!];
+const lesson = chapter.lessons[chapter?.firstLessonId!];
```
SamVerschueren, line 8: I'm very confused. So all the `firstPartId` and `firstChapterId` and `firstLessonId` are defined as optional in `entities/index.ts`. Why are we here telling TS that they will be defined?

Also, what if they are defined but they don't exist? Then `chapter` will be undefined and it will throw, no?

SamVerschueren (commented):
It's a bit hard to follow for me, but maybe it is because I don't really know how it worked before (except that it was based on the folder name).

packages/astro/src/default/utils/content.ts
```diff
@@ -180,32 +304,15 @@ function sortCollection(collection: CollectionEntryTutorial[]) {
...
-      const numB = parseInt(splitB[i], 10);
-
-      if (!isNaN(numA) && !isNaN(numB) && numA !== numB) {
-        return numA - numB;
-      } else {
-        if (splitA[i] !== splitB[i]) {
-          return splitA[i].localeCompare(splitB[i]);
-        }
+      if (splitA[i] !== splitB[i]) {
+        return splitA[i].localeCompare(splitB[i]);
```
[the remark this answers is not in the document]
Nemikolh: Actually I'll remove that entire for loop. It's not needed anymore and it's confusing.

packages/astro/src/default/pages/index.astro
Nemikolh: The folder were expected to start at `1`. We now instead pick the first folder that comes up. It could start at anything: 0, 4, ... or be explicitly defined. 

So this is a new tiny feature! :smiley:

packages/astro/src/default/utils/content.ts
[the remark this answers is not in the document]
SamVerschueren: So the current behaviour still works then?

SamVerschueren (approved):

SamVerschueren (approved):
Great work!

PR #81, opened 2024-06-19
feat: `tutorialkit eject` command

Nemikolh:
This PR adds an `eject` command to the `tutorialkit` CLI so that you can get back full control if you ever need to.

It works by modifying the `astro.config.ts` file to turn off the `defaultRoutes` and then copy the content of `@tutorialkit/astro/dist/default` into the `src` folder of the TutorialKit project.

I renamed `env.d.ts` in `astro/src/default/` to `env-default.d.ts` so that they are no collision as they have different content. Ideally we should merge them but having two separate file is simpler to start with.

Note that `--force` will be required because of the `env.d.ts` for TutorialKit project that have already be generated. The collision will only be avoided for future projects.


packages/cli/src/commands/eject/index.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  if (folderPath === undefined) {
+    folderPath = process.cwd();
+  } else {
+    folderPath = path.resolve(process.cwd(), folderPath);
+  }
+
+  /**
+   * First we make sure that the destination has the correct files
+   * and that there won't be any files overriden in the process.
```
d3lm, line 52:
```suggestion
   * and that there won't be any files overwritten in the process.
```

packages/cli/src/commands/eject/index.ts
```diff
@@ -0,0 +1,132 @@
...
+   *
+   * If they are any and `force` was not specified we abort.
```
d3lm, line 54:
```suggestion
   * If there are any and `force` was not specified we abort.
```

d3lm (commented):
Cool 👏. Minor remarks.

packages/cli/src/index.ts
```diff
@@ -3,12 +3,13 @@
...
 import { createTutorial } from './commands/create/index.js';
+import { ejectRoutes } from './commands/eject/index.js';
 import { pkg } from './pkg.js';
 import { errorLabel, primaryLabel, printHelp } from './utils/messages.js';
 
-type CLICommand = 'version' | 'help' | 'create';
+type CLICommand = 'version' | 'help' | 'create' | 'eject';
 
-const supportedCommands = new Set(['version', 'help', 'create']);
+const supportedCommands = new Set<string>(['version', 'help', 'create', 'eject'] satisfies CLICommand[]);
```
SamVerschueren, line 12: Can't we just type this as `Set<CLICommand>`?

packages/cli/src/commands/eject/index.ts
```diff
@@ -0,0 +1,132 @@
...
+  const srcPath = path.join(localAstroIntegration, 'dist', 'default');
+
+  // check that they are no collision
+  if (!force) {
+    walk(srcPath, (relativePath) => {
+      const destination = path.join(srcDestPath, relativePath);
+
+      if (fs.existsSync(destination)) {
+        throw new Error(
+          `Eject aborted because '${destination}' would be overriden by this command. Use --force to ignore this error.`,
```
SamVerschueren, line 94:
```suggestion
          `Eject aborted because '${destination}' would be overwritten by this command. Use --force to ignore this error.`,
```

packages/cli/src/utils/astro-config.ts
```diff
@@ -156,9 +167,9 @@ function updateObject(properties: any, object: t.ObjectExpression | undefined):
 
   object ??= t.objectExpression([]);
 
-  for (const property of properties) {
+  for (const property in properties) {
```
SamVerschueren, line 170: Do we need to use `Object.hasOwn()` here to make sure that the property is not coming from some prototype? I don't know if that's possible in this case though.

```ts
const bar = {
    baz: 'hello',
};

//////////////////////////

const obj = {
    foo: 'bar',
};

Object.setPrototypeOf(obj, bar);

for (const key in obj) {
    if (Object.hasOwn(obj, key)) {
        console.log(key);
    }
}
```

Without `Object.hasOwn()`, it will print both `foo` and `baz`. Now it only prints `foo`.

packages/cli/src/commands/eject/index.ts
```diff
@@ -0,0 +1,132 @@
...
+  assertExists(folder);
+
+  const astroConfigPath = assertExists(path.join(folder, 'astro.config.ts'));
+  const srcDestPath = assertExists(path.join(folder, 'src'));
+
+  const localAstroIntegration = assertExists(path.resolve(folder, 'node_modules', '@tutorialkit', 'astro'));
+
+  const srcPath = path.join(localAstroIntegration, 'dist', 'default');
+
+  // check that they are no collision
```
SamVerschueren, line 87:
```suggestion
  // check that there are no collision
```

SamVerschueren (commented):
Added some comments and questions.

Can we add a test for this as well? I think it should be pretty straight forward in the `create-tutorial.test.ts` by creating the project, ejecting, and taking a snapshot.

packages/cli/src/index.ts
Nemikolh: We can't because then we get a type error below when trying to do `supportedCommands.has(string)`. 

The reason I made that change was to make sure the values are checked against the `CLICommand` type.

packages/cli/src/utils/astro-config.ts
Nemikolh: `properties` is only intended to be a POJO here and is only used with an object literal that we created (it's not a function used with a value produced by some code out of our control).

So that means that the check is not needed.

Happy to add it if we think it could be useful in the future though :+1:

SamVerschueren: No all good if we are sure it's not causing any harm 👍 .

packages/cli/src/commands/eject/index.ts
```diff
@@ -0,0 +1,175 @@
...
+  const pkgJson: PackageJson = JSON.parse(fs.readFileSync(pkgJsonPath, 'utf-8'));
+  const astroIntegrationPkgJson: PackageJson = JSON.parse(
+    fs.readFileSync(path.join(astroIntegrationPath, 'package.json'), 'utf-8'),
+  );
+
+  let hasChanged = false;
+
+  for (const dep of REQUIRED_DEPENDENCIES) {
+    if (!(dep in pkgJson.dependencies) && !(dep in pkgJson.devDependencies)) {
+      pkgJson.dependencies[dep] = astroIntegrationPkgJson.dependencies[dep];
```
SamVerschueren, line 100: Maybe not super important for now, but we might want to check the version of the dependency and warn the user that they might need to upgrade/downgrade or something.

packages/cli/src/commands/eject/index.ts
```diff
@@ -0,0 +1,175 @@
...
+
+      hasChanged = true;
+    }
+  }
+
+  updateWorkspaceVersions(pkgJson.dependencies, TUTORIALKIT_VERSION, (dependency) =>
```
SamVerschueren, line 106: Why would that package.json have workspace dependencies? I'm a little confused :p.

Nemikolh: Good question! It's only for when running the tests. It's actually the same usecase than when we call that function in `create`.

packages/cli/tests/create-tutorial.test.ts
```diff
@@ -69,6 +69,66 @@ test('create and build a project', async (context) => {
   expect(distFiles.map(normaliseSlash).sort()).toMatchSnapshot();
 });
 
+test('create and eject a project', async (context) => {
```
SamVerschueren, line 72: Maybe also try to eject from something that is not an TK project?

packages/cli/src/commands/eject/index.ts
Nemikolh: We do! The `let hasChanged = false;` is used for that purpose.

packages/cli/src/commands/eject/index.ts
```diff
@@ -0,0 +1,175 @@
...
+    REQUIRED_DEPENDENCIES.includes(dependency),
+  );
+
+  if (hasChanged) {
+    fs.writeFileSync(pkgJsonPath, JSON.stringify(pkgJson, undefined, 2), { encoding: 'utf-8' });
+
+    console.log(primaryLabel('INFO'), 'New dependencies added, install the new dependencies before proceeding');
```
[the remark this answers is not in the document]
Nemikolh: We could indeed to it for them but I kept it simple for now.

If there are no `node_modules` then the `eject` command won't work because it moves the files of the version of `@tutorialkit/astro` that they use (instead of using a possibily incompatible one).

packages/cli/tests/create-tutorial.test.ts
Nemikolh: Ah that's a very good point! I should write a test for that too :+1:

Nemikolh: Tests added :+1:

packages/cli/src/commands/eject/index.ts
```diff
@@ -0,0 +1,175 @@
...
+   */
+  const astroConfig = await parseAstroConfig(astroConfigPath);
+
+  replaceArgs({ defaultRoutes: false }, astroConfig);
+
+  fs.writeFileSync(astroConfigPath, generateAstroConfig(astroConfig));
+
+  /**
+   * We copy all assets from the `default` folder into the `src` folder.
+   */
```
d3lm, line 84: Nit: Since this is a one liner we could make that a line comment:
```suggestion
// we copy all assets from the `default` folder into the `src` folder
```

packages/cli/src/commands/eject/index.ts
```diff
@@ -0,0 +1,175 @@
...
+  fs.cpSync(srcPath, srcDestPath, { recursive: true });
+
+  /**
+   * Last, we ensure that the package.json contains the extra dependencies.
+   * If any are missing we suggest to install the new dependencies.
+   */
+  const pkgJson: PackageJson = JSON.parse(fs.readFileSync(pkgJsonPath, 'utf-8'));
+  const astroIntegrationPkgJson: PackageJson = JSON.parse(
```
d3lm, line 92: Nit: Newline above since the `JSON.parse` spans over multiple lines.

packages/cli/src/commands/eject/index.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  fs.writeFileSync(astroConfigPath, generateAstroConfig(astroConfig));
+
+  /**
+   * We copy all assets from the `default` folder into the `src` folder.
+   */
+  fs.cpSync(srcPath, srcDestPath, { recursive: true });
+
+  /**
+   * Last, we ensure that the package.json contains the extra dependencies.
```
d3lm, line 88:
```suggestion
   * Last, we ensure that the `package.json` contains the extra dependencies.
```

packages/cli/src/commands/eject/index.ts
d3lm: Hm so technically this can result in false positives if say I have `nanostores` in my `package.json` already but the range just doesn't match but the version would be the same, e.g.
```
^1.1.1 !== 1.1.1
```
Given that there's no new patch version and `1.1.1` is the latest. Wouldn't it be better to check node_modules or check the lock file? But maybe this is good enough and a warning is good in any case which would force the user to double check the dependency.

packages/cli/src/commands/eject/index.ts
d3lm, line 113: Maybe we should list the ones that were updated / changed just for transparency? Similar to how Astro CLI and other CLIs show you the modifications they do to files.

packages/cli/src/commands/eject/index.ts
```diff
@@ -0,0 +1,175 @@
...
+  const srcPath = path.join(astroIntegrationPath, 'dist', 'default');
+
+  // check that there are no collision
+  if (!force) {
+    walk(srcPath, (relativePath) => {
+      const destination = path.join(srcDestPath, relativePath);
+
+      if (fs.existsSync(destination)) {
+        throw new Error(
+          `Eject aborted because '${destination}' would be overwritten by this command. Use --force to ignore this error.`,
```
d3lm, line 135: Maybe print `--force` in yellow text to draw some attention to that flag?

packages/cli/tests/create-tutorial.test.ts
```diff
@@ -69,6 +69,164 @@ test('create and build a project', async (context) => {
...
+  const dest = path.join(tmpDir, name);
+
+  await fs.mkdir(dest);
+  await fs.mkdir(path.join(dest, 'src'));
+
+  await fs.writeFile(
+    path.join(dest, 'package.json'),
+    JSON.stringify({ name: 'astro', dependencies: { astro: '4.11.0' } }),
+  );
+  await fs.writeFile(
```
d3lm, line 177: Newline before.

packages/cli/tests/create-tutorial.test.ts
```diff
@@ -69,6 +69,164 @@ test('create and build a project', async (context) => {
...
+  const dest = path.join(tmpDir, name);
+
+  await fs.mkdir(dest);
+  await fs.mkdir(path.join(dest, 'src'));
+
+  await fs.writeFile(
+    path.join(dest, 'package.json'),
+    JSON.stringify({ name: 'astro', dependencies: { astro: '4.11.0', '@tutorialkit/astro': '*' } }),
+  );
+  await fs.writeFile(
```
d3lm, line 207: Newline before.

packages/cli/tests/create-tutorial.test.ts
d3lm, line 1: Nice tests!

Should we also snapshot the astro config that we eject to make sure that the content is correct etc.? Maybe not be as valuable because we also run the build and if it works maybe that's already a good enough proof but snapshotting the config is just another test to prevent potential regressions maybe.

d3lm (commented):
Nice one. Left some minor comments.

packages/cli/src/commands/eject/index.ts
Nemikolh:
```suggestion

  const astroIntegrationPkgJson: PackageJson = JSON.parse(
```

packages/cli/tests/create-tutorial.test.ts
Nemikolh:
```suggestion

  await fs.writeFile(
```

packages/cli/tests/create-tutorial.test.ts
Nemikolh:
```suggestion

  await fs.writeFile(
```

packages/cli/src/commands/eject/index.ts
Nemikolh: If they have an incompatible version of `nanostores`, there's not much we can do as they'll have to update the code they've just ejected. Hopefully there's a TypeScript error so that they know.

If it's compatible, then this code uses their version, so I think this code is good?

packages/cli/tests/create-tutorial.test.ts
Nemikolh: Ah good point! I'll add it to the `create and eject a project` test :+1:

packages/cli/src/commands/eject/index.ts
d3lm: Yea I think it's fine cause we only show the warning if that dependency is missing entirely in the `package.json`. I was just wondering if we can be smarter here and maybe check the version and print a hint if the version doesn't match to tell them that there may be incompatibilities.

Nemikolh: Ah good point! Oh yeah that would be really nice actually.

d3lm: One possible way would be to leverage `npm why --json` to see what version is installed and what version we require.

packages/cli/src/commands/eject/index.ts
```diff
@@ -0,0 +1,177 @@
...
+      newDependencies.push(dep);
+    }
+  }
+
+  updateWorkspaceVersions(pkgJson.dependencies, TUTORIALKIT_VERSION, (dependency) =>
+    REQUIRED_DEPENDENCIES.includes(dependency),
+  );
+
+  if (newDependencies.length > 0) {
+    fs.writeFileSync(pkgJsonPath, JSON.stringify(pkgJson, undefined, 2), { encoding: 'utf-8' });
```
SamVerschueren, line 110: I wonder if we should use [detect-indent](https://www.npmjs.com/package/detect-indent) and use the correct indentation when rewriting the JSON file.

Back in the day, `npm install` always used 2 spaces and a lot of people where actually annoyed by it. Nowadays they also detect the indentation and rewrite the file correctly based on what it already had.

packages/cli/src/commands/eject/index.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    console.log(
+      primaryLabel('INFO'),
+      `New dependencies added: ${newDependencies.join(', ')}. Install the new dependencies before proceeding`,
+    );
```
SamVerschueren, line 115: Should we ask to re-install the deps and do it for them?

SamVerschueren (approved):
Awesome work on this @Nemikolh 🙏 !

packages/cli/src/commands/eject/index.ts
Nemikolh: Oh that's a really good idea! Didn't know about that package, love it!!

packages/cli/src/commands/eject/index.ts
Nemikolh: Oh I like it, we can even use [`preferred-pm`](https://www.npmjs.com/package/preferred-pm) to know which one to use :raised_hands:

Nemikolh: Updated :raised_hands:

packages/cli/src/commands/eject/index.ts
Nemikolh: Updated :raised_hands:

packages/cli/package.json
d3lm, line 1: One thing to keep in mind with adding more packages is that the CLI becomes heavier and we still want it to work well on SB so we should keep an eye on the number of dependencies because all those need to be installed if you run `npm create tutorial` on StackBlitz.com.

packages/cli/src/commands/eject/index.ts
```diff
@@ -0,0 +1,196 @@
...
+  updateWorkspaceVersions(pkgJson.dependencies, TUTORIALKIT_VERSION, (dependency) =>
+    REQUIRED_DEPENDENCIES.includes(dependency),
+  );
+
+  if (newDependencies.length > 0) {
+    fs.writeFileSync(pkgJsonPath, JSON.stringify(pkgJson, undefined, indent), { encoding: 'utf-8' });
+
+    console.log(
+      primaryLabel('INFO'),
+      `New dependencies added: ${newDependencies.join(', ')}. Install the new dependencies before proceeding`,
```
d3lm, line 121: Tiny nit:
```suggestion
      `New dependencies added: ${newDependencies.join(', ')}. Install the new dependencies before proceeding.`,
```

packages/cli/src/commands/eject/index.ts
```diff
@@ -0,0 +1,196 @@
...
+    );
+
+    if (!flags.defaults) {
+      const packageManager = (await preferredPM(pkgJsonPath)).name;
+
+      const answer = await prompts.confirm({
+        message: `Do you want to install those dependencies now using ${packageManager}?`,
```
d3lm, line 128: Nit: Shall we colorize the preferred package manager?

d3lm (approved):
👏

packages/cli/src/commands/eject/index.ts
Nemikolh: Oh good idea! I'll play around with everything we have so far to see which color would be the best :+1:

packages/cli/package.json
Nemikolh: Very good point, so `detect-indent` has no dependencies so we're good on that one. However `preferred-pm` has a lot:

```
preferred-pm 3.1.3
├─┬ find-up 5.0.0
│ ├─┬ locate-path 6.0.0
│ │ └─┬ p-locate 5.0.0
│ │   └─┬ p-limit 3.1.0
│ │     └── yocto-queue 0.1.0
│ └── path-exists 4.0.0
├─┬ find-yarn-workspace-root2 1.2.16
│ ├─┬ micromatch 4.0.7
│ │ ├─┬ braces 3.0.3
│ │ │ └─┬ fill-range 7.1.1
│ │ │   └─┬ to-regex-range 5.0.1
│ │ │     └── is-number 7.0.0
│ │ └── picomatch 2.3.1
│ └─┬ pkg-dir 4.2.0
│   └─┬ find-up 4.1.0
│     ├─┬ locate-path 5.0.0
│     │ └─┬ p-locate 4.1.0
│     │   └─┬ p-limit 2.3.0
│     │     └── p-try 2.2.0
│     └── path-exists 4.0.0
├── path-exists 4.0.0
└─┬ which-pm 2.0.0
  ├─┬ load-yaml-file 0.2.0
  │ ├── graceful-fs 4.2.11
  │ ├─┬ js-yaml 3.14.1
  │ │ ├─┬ argparse 1.0.10
  │ │ │ └── sprintf-js 1.0.3
  │ │ └── esprima 4.0.1
  │ ├── pify 4.0.1
  │ └── strip-bom 3.0.0
  └── path-exists 4.0.0
```

Going to https://pkg-size.dev/preferred-pm, it's 1.2MB (29 packages). However I think we can restrict ourselves to using `which-pm` from that graph which is only 810KB: https://pkg-size.dev/which-pm

PR #82, opened 2024-06-19
chore: fix issue with changelog generation

Nemikolh:
This PR fixes an issue where packages `CHANGELOG.md` files included changes unrelated to them.


scripts/changelog.mjs
```diff
@@ -104,7 +104,7 @@ function generateChangelog(pkg) {
   };
 
   const gitRawCommitsOpts = {
-    path: pkg.gitPath ?? path.dirname(pkg.path),
+    path: pkg.gitPath ?? pkg.path,
```
Nemikolh, line 107: This was the source of the bug :sweat_smile:

d3lm (approved):
Cool. Would be super sweet if we can improve the formatting in terms of the extra newlines but if it's too much work then maybe it's not worth it. My inner [Monk](https://en.wikipedia.org/wiki/Monk_(TV_series)) tells me to fix this 🙉

Nemikolh:
Ah sorry I forgot about that one! I'll look into using our own preset 👍

SamVerschueren (approved):

d3lm:
Any updates on this one?

> (Btw, Monk is such a good series 🤩)

Yes! Too good.

Nemikolh:
@d3lm So we had talked about this PR with @SamVerschueren and decided to wait for you to be back before moving forward with this PR.

In particular we were thinking that it could be great to also auto-include the contributors to give them a big shoutout as part of the changelog.

@SamVerschueren had an idea of using the GitHub API in order to get all the commits that should be included in the changelog and also get the Contributors usernames. If we went that route though it essentially mean writing our own tool which I think is fine but meant that we should probably discuss it first.

d3lm:
Yea I like auto-adding the contributors to the changelog!

Only thing I am wondering about is value and effort. I think we have plenty of other high prio things to work. I generally like the idea that @SamVerschueren had but since we don't have a ton of contributors yet the investment may not be worth it right now, wdyt? What I am thinking is that we can start automating this once we have more external people contributing to the project.

Nemikolh:
Yeah, I agree. Let's see what Sam says 👍

SamVerschueren:
Yeah I totally agree. Let's postpone this.

PR #84, opened 2024-06-20
chore: update dependencies

SamVerschueren:
I used `pnpm up -r --workspace` to update all the dependencies to the latest version based on the semver range defined in the `package.json` files.

The main reason was that I'm moving `demo.tutorialkit.dev` into this repository and wanted to use the workspace tutorialkit packages. But I got a mismatch in react versions because of that. So I figured it might be better to first separately update all the versions.


Nemikolh (approved):
Nice! Thanks for doing it!

PR #83, opened 2024-06-20
docs: add demo.tutorialkit.dev website

SamVerschueren:
This PR brings the source of https://demo.tutorialkit.dev to this repository to make it easier for external contributors to contribute.

I updated the dependencies to use the workspace dependencies instead. This means that the demo is now always build with the version of TK that is on `main`.

PR #85, opened 2024-06-20
ci: build docs

SamVerschueren:
This PR adds an additional job to the CI workflow.

The idea is to also make sure that our docs still build. It's probably less of an issue for the docs website itself, but I also want to use the tutorialkit workspace packages for the demo (see #83). This means that whatever changes we make in our packages, we still make sure that our demo website keeps working.


Nemikolh (approved):
Pretty cool!

PR #87, opened 2024-06-20
feat(theme): create proper tokens for callouts

SamVerschueren:
I'm documenting the theming aspect of TutorialKit and noticed that it's pretty hard to change the colors of the callouts specifically.

That's why I extracted them per type into `variables.css` and referenced them in `markdown.css`. This way, it's easier for a user to overwrite specific colors.

One thing I wasn't sure about was if I needed to add them to `theme.ts`? The variables are not used by uno-css (yet) because we just reference them in `markdown.css`. But should I already add them?

I also noticed that there are even more things in `markdown.css` which we might want to expose later in regards to styling codeblocks and what not. But I left that out for now.


Nemikolh (approved):

donmckenna:
I think Dom would know better regarding the uno-css question, but otherwise lgtm ❤️

d3lm (approved):

d3lm:
Yep looks good. We don't need to define them in `theme.ts` since we don't use those more specific variables as CSS classes 👍

PR #90, opened 2024-06-24
feat: add eslint

d3lm:
This PR adds our eslint plugin to enforce our coding convention across the monorepo.

I have already fixed all linting issues that were reported and streamlined some of the CSS files as well to match our conventions. We should definitely add some style (CSS, SCSS) linting as well at some point.

ℹ️ ~~Note that running this won't work as of right now because the new version of the eslint plugin is not yet published.~~ ✅


d3lm:
Updated the PR to use the alpha version that was released.

package.json
```diff
@@ -11,11 +11,13 @@
     "docs:build": "pnpm run --filter=tutorialkit.dev build",
     "demo": "pnpm run --filter=demo.tutorialkit.dev dev",
     "demo:build": "pnpm run --filter=demo.tutorialkit.dev build",
+    "lint": "eslint \"{packages,docs}/**/*.{ts,tsx,js,json,jsonc,js,mjs,cjs}\"",
```
SamVerschueren, line 14: Is eslint covering prettier as well?

Any reason we don't include all files and folders?

SamVerschueren (approved):
Great work @d3lm 🙏 . Love to see the consistency!

package.json
d3lm: Yes, our eslint plugin always included the eslint recommended rules and it also takes the prettier config into account.

> Any reason we don't include all files and folders?

I only did this to reduce the number of files to be checked and to be more specific. Other than that there's no particular reason. Cause those extensions are the only files that our plugin covers, but I think it would be fine to remove the extensions since the plugin contains the file extensions too and it would ignore other files.

packages/cli/src/commands/create/astro-config.ts
```diff
@@ -21,14 +21,14 @@ export async function parseAstroConfig(astroConfigPath: string): Promise<t.File>
...
- * For instance if `tutorialkit` is currently invoked as:
+ * For example:
  *
  * ```ts
  * tutorialkit({ isolation: 'require-corp' })
  * ```
  *
- * and this function is called with `{ enterprise: {} }` as the new tutorialkit object, then
- * the modified config will contain:
+ * If `tutorialkit` is currently invoked as, and this function is called with `{ enterprise: {} }`
```
Nemikolh, line 30: Nit:

```suggestion
 * If `tutorialkit` is currently invoked like in the above example, and this function is called with `{ enterprise: {} }`
```

packages/cli/src/commands/create/astro-config.ts
```diff
@@ -171,6 +171,8 @@ function updateObject(properties: any, object: t.ObjectExpression | undefined):
       }
     }
   }
+
+  return object;
```
Nemikolh, line 175: Oh interesting, this just made visible that the return value is not used. We should probably have line `154` just do a `return`.

packages/astro/src/default/components/webcontainer.ts
```diff
@@ -1,14 +1,16 @@
-import { useAuth } from './setup.js';
-import { auth, WebContainer } from '@webcontainer/api';
 import { TutorialStore } from '@tutorialkit/runtime';
+import { auth, WebContainer } from '@webcontainer/api';
+import { useAuth } from './setup.js';
```
Nemikolh, line 3: Oh this import should always be first as it sets a global read by the `@webcontainer/api` package.

packages/runtime/src/tutorial-runner.spec.ts
```diff
@@ -1,11 +1,11 @@
-import { resetProcessFactory, setProcessFactory } from 'test-utils';
 import { WebContainer } from '@webcontainer/api';
-import { describe, test, expect, beforeEach, vi } from 'vitest';
 import type { MockedWebContainer } from 'test-utils';
-import { TutorialRunner } from './tutorial-runner.js';
-import { StepsController } from './webcontainer/steps.js';
+import { resetProcessFactory, setProcessFactory } from 'test-utils';
```
Nemikolh, line 3: This import must be first in order for the mock to be installed and mock the `@webcontainer/api` package.

Nemikolh (approved):
Looks pretty good!

packages/cli/src/commands/create/astro-config.ts
d3lm: Yep I was wondering the same cause the return value is not used but I wasn't sure and I didn't wanna change it too much.

d3lm: Fixed it.

packages/runtime/src/tutorial-runner.spec.ts
d3lm: Yep good point. Fixed this!

packages/cli/src/commands/create/astro-config.ts
Nemikolh: Yay!! :star_struck:

Nemikolh (approved):
This is amazing!! Really good work both here and on the eslint plugin @d3lm 

:tada: :tada:

PR #91, opened 2024-06-24
feat: mobile support

Nemikolh:
This PR adds mobile support for TutorialKit following a similar approach as the excellent https://learn.svelte.dev/.

See the screencast below to see how it looks:

[Screencast from 2024-06-24 14-58-32.webm](https://github.com/stackblitz/tutorialkit/assets/4008156/5fc3faf3-f0b1-48b7-86fd-d17d3e988cb5)


packages/astro/src/default/components/ResizablePanel.astro
```diff
@@ -208,89 +233,3 @@ if (!hasSidePanel) {
 
   customElements.define('resizable-panel', ResizablePanel);
 </script>
-
-<style>
```
Nemikolh, line 212: I replaced the inline style from this component with unocss classes as it makes it easier to add breakpoints for mobile.

packages/runtime/src/webcontainer/on-demand-boot.ts
```diff
@@ -0,0 +1,50 @@
...
+let blocked: undefined | boolean;
+const blockingStatus = withResolvers<void>();
+
+export async function safeBoot(options: BootOptions) {
+  if (typeof blocked === 'undefined') {
+    blocked = isRestricted();
+  }
+
+  if (blocked) {
+    await blockingStatus.promise;
```
Nemikolh, line 18: We now prevent WebContainer from booting on iOS to avoid crashing the page as it sometimes does. Instead the idea is to later let them proceeed with the boot by pressing a button.

packages/components/react/src/Panels/PreviewPanel.tsx
```diff
@@ -61,6 +61,13 @@ export const PreviewPanel = memo(
       [],
     );
 
+    // we update the iframes position at ~30fps, needed for mobile
+    useEffect(() => {
+      const intervalId = setInterval(onResize, 20);
+
+      return () => clearInterval(intervalId);
+    }, []);
```
Nemikolh, line 69: With this code we make sure that the preview are correctly positioned. This is important on mobile when switching between the content and the editor.

I kept the existing `ResizeObserver` instances because they are smoother. 

This also fix the edge case where the preview stop moving when increasing the size of the terminal section.

packages/astro/src/default/layouts/Layout.astro
```diff
@@ -49,6 +49,9 @@ const { title } = Astro.props;
 
           if (newMain && oldMain) {
             builtInSwap.swapBodyElement(newMain, oldMain);
+
+            // delete extraneous route announcer
+            document.querySelector('.astro-route-announcer')?.remove();
```
Nemikolh, line 54: Unrelated fix to this PR but I noticed that as you navigate between pages we were accumulating `div`s under the body.

packages/astro/src/default/components/MobileContentToggle.astro
```diff
@@ -0,0 +1,50 @@
+---
+
+---
```
SamVerschueren, line 3: Is this needed even though it's empty?

packages/astro/src/default/components/MobileContentToggle.astro
```diff
@@ -0,0 +1,50 @@
...
+
+      const [tutorialBtn, toggleButton, editorButton] = this.querySelectorAll(
+        ':scope > button',
+      ) as NodeListOf<HTMLButtonElement>;
+
+      const toggle = toggleButton.querySelector('span')!;
+
+      tutorialBtn.onclick = () => setView('content');
+      editorButton.onclick = () => setView('editor');
+      toggleButton.onclick = () => setView(viewStore.get() === 'content' ? 'editor' : 'content');
```
SamVerschueren, line 31: Is a touch identical to a click?

packages/components/react/src/Panels/PreviewPanel.tsx
```diff
@@ -61,6 +61,13 @@ export const PreviewPanel = memo(
       [],
     );
 
+    // we update the iframes position at ~30fps, needed for mobile
+    useEffect(() => {
+      const intervalId = setInterval(onResize, 20);
```
SamVerschueren, line 66: We don't want to use `requestAnimationFrame` instead? Or would that be too much?

packages/astro/src/default/components/ResizablePanel.astro
```diff
@@ -111,6 +125,18 @@ if (!hasSidePanel) {
       }
     }
 
+    mainPanel(): HTMLElement {
+      return this.#container.children[1] as HTMLElement;
```
SamVerschueren, line 129: It feels a bit tricky to rely on the position? Can't we use a data-element or something and grab the correct child?

packages/components/react/src/Panels/PreviewPanel.tsx
Nemikolh: Good question, I'm torn on that one and I'm not sure. I feel like it would be too much, however if we did that then we can discard the `ResizeObserver` logic entirely which is also pretty neat.

SamVerschueren (commented):
Did a first initial review and added some questions. Looks very cool btw!

packages/astro/src/default/components/ResizablePanel.astro
Nemikolh: Oh yeah, I'll update it :+1: 

(I did it this way only because the template is above so it's easy to check if it's correct. But if this code ever grows then it won't be as maintainable that `data-id`s)

packages/astro/src/default/components/MobileContentToggle.astro
Nemikolh: Yes! It's actually even more than [a mouse click](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event):

> An element receives a click event when any of the following occurs:
>
> * a pointing-device button (such as a mouse's primary button) is both pressed and released while the pointer is located inside the element.
> * a touch gesture is performed on the element
> * the Space key or Enter key is pressed while the element is focused

packages/astro/src/default/components/MobileContentToggle.astro
Nemikolh: Oh you're right, I think it isn't! :open_mouth:

packages/astro/src/default/components/MobileContentToggle.astro
```diff
@@ -0,0 +1,50 @@
...
+<script>
+  import { viewStore, type View } from '../stores/view-store';
+
+  class ViewToggle extends HTMLElement {
+    constructor() {
+      super();
+
+      const [tutorialBtn, toggleButton, editorButton] = this.querySelectorAll(
+        ':scope > button',
+      ) as NodeListOf<HTMLButtonElement>;
```
AriPerkkio, line 25: Tip: `querySelectorAll` has good type narrowing. Instead of using type assertion on it's output, we could use type assertion on the input. It's much better to say `':scope > button'` selector captures `buttons`, than to say that `querySelectorAll` returns `NodeListOf<HTMLButtonElement>`.

```suggestion
      const [tutorialBtn, toggleButton, editorButton] = this.querySelectorAll(
        ':scope > button' as 'button',
      );
```

packages/astro/src/default/components/MobileContentToggle.astro
```diff
@@ -0,0 +1,50 @@
...
+<div class="h-12 sm:hidden"></div>
+<view-toggle
+  class="fixed sm:hidden h-12 bottom-0 w-full bg-tk-elements-app-backgroundColor border-t border-tk-elements-app-borderColor flex justify-center items-center z-60 text-tk-elements-app-textColor"
+>
+  <button aria-hidden="true">Tutorial</button>
+  <button aria-pressed="true" class="rounded-full w-8 h-4 p-0.5 bg-gray-4 dark:bg-gray-6 mx-2 relative">
+    <span class="inline-block transition-all absolute top-0.5 left-0.5 rounded-full bg-white dark:bg-gray-2 w-3 h-3"
+    ></span>
+  </button>
+  <button aria-hidden="true">Editor</button>
```
AriPerkkio, line 14: I would recommend to remove the `aria-hidden`'s from here. Right now these elements are present in tab order but excluded from accessibility tree. Assistive technologies will not read these but users will still be able to navigate there. Also the 2nd button is present in accessibility tree but doesn't have any name so it's announced just as button. And as the `aria-pressed` state is not updated, maybe remove that one too. 

There are some accessibility issues in the project at the moment so might be good to tackle all of those at once at some point.

packages/astro/src/default/components/MobileContentToggle.astro
Nemikolh: Oh nice! I didn't knew about that trick, really cool :smiley:

packages/astro/src/default/components/MobileContentToggle.astro
Nemikolh: Yep makes sense, thanks for the feedback :pray:

Nemikolh:
@d3lm It's updated now!

packages/astro/src/default/components/MobileContentToggle.astro
```diff
@@ -0,0 +1,44 @@
...
+  <button>Editor</button>
+</view-toggle>
+<script>
+  import { viewStore, type View } from '../stores/view-store';
+
+  class ViewToggle extends HTMLElement {
+    constructor() {
+      super();
+
+      const [tutorialBtn, toggleButton, editorButton] = this.querySelectorAll(':scope > button' as 'button');
```
SamVerschueren, line 19: Why is the `:scope > button` casted to the string `button`? Does that have any benefit?

packages/astro/src/default/components/ResizablePanel.astro
```diff
@@ -23,18 +37,16 @@ if (!hasSidePanel) {
...
-  data-pos={pos}
-  data-min={min}
-  data-priority={priority}
-  data-max={max}
->
-  <div class=`container max-w-full ${type}` style=`--pos: ${pos}`>
+<resizable-panel class={className} data-id={id} data-type={type} data-pos={pos} data-min={min} data-max={max}>
+  <div
+    data-id="container"
+    class=`sm:grid relative w-full h-full max-w-full ${type === 'horizontal' ? 'sm:grid-cols-[var(--pos)_1fr]' : 'sm:grid-rows-[var(--pos)_1fr]'}`
```
SamVerschueren, line 47: Why not use the `classNames()` method here?

packages/astro/src/default/components/ResizablePanel.astro
```diff
@@ -44,36 +56,38 @@ if (!hasSidePanel) {
...
       hasSidePanel && (
         <>
-          <div class="panel">
+          <div class={`${panelClass} ${sidePanelClass}`} data-id="side-panel">
             <slot name="b" />
           </div>
-          <div class="divider" />
+          <div
+            data-id="divider"
+            class={`absolute z-90 transition-colors hover:bg-gray-500/13 ${type === 'horizontal' ? 'w-0 h-full left-[var(--pos)] cursor-ew-resize p-0 px-1.5 -translate-x-1/2' : 'w-full h-0 top-[var(--pos)] cursor-ns-resize p-0 py-2 -translate-y-1/2'}`}
```
SamVerschueren, line 77: Same here, might be more readable with `classNames()`?

packages/astro/src/default/components/MobileContentToggle.astro
Nemikolh: It does! :smiley:  That was a suggestion from Ari: https://github.com/stackblitz/tutorialkit/pull/91#discussion_r1652969703

packages/components/react/src/Panels/PreviewPanel.tsx
SamVerschueren: I think I would prefer/try to use `requestAnimationFrame`. We could use the callback argument to calculate it so that it will resize at approximately 60 fps if we wanted that (or 30) https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame#examples. But if everyone is ok to just use `setInterval()`, I'm fine with it as well though.

packages/astro/src/default/components/ResizablePanel.astro
Nemikolh: Oh good call! Will update it :+1:

SamVerschueren (approved):
Looks pretty nice!

SamVerschueren:
Is it possible that this is broken now on desktop? If I open the preview, I only see the editor view and I can't toggle to the content either.

https://4c6936c9.tutorialkit-demo-page.pages.dev/forms-tutorial/introduction/welcome/

Nemikolh:
@SamVerschueren Oh wow, maybe a merge commit introduced a bug. I'll address all the feedback and test it again.

packages/astro/src/default/components/MobileContentToggle.astro
SamVerschueren: TIL!

packages/astro/src/default/components/MobileContentToggle.astro
d3lm: Yea that's the power of string literal types 🙌 🔥

packages/astro/src/default/components/ResizablePanel.astro
d3lm: Yep, would be good to use `classNames` (I find it a lot easier to read personally).

packages/components/react/src/Panels/PreviewPanel.tsx
d3lm: Yea I'd also prefer `requestAnimationFrame` so we can use the max number of frames that the browser allows us to render at.

packages/runtime/src/webcontainer/on-demand-boot.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * Lightweight wrapper around WebContainer.boot
```
d3lm, line 2:
```suggestion
 * Lightweight wrapper around `WebContainer.boot`
```

d3lm: This comment is a bit hard to read and could benefit from a slight rewording.
```
Lightweight wrapper around "WebContainer.boot" to only boot if requested on iOS, since booting might risk running out of memory.
```

packages/runtime/src/webcontainer/on-demand-boot.ts
```diff
@@ -0,0 +1,50 @@
...
+ * to only boot if requested on iOS as typically
+ * boot might risk the page to run out of memory.
+ */
+import { WebContainer, type BootOptions } from '@webcontainer/api';
+import { withResolvers } from '../utils/promises.js';
+
+let blocked: undefined | boolean;
+const blockingStatus = withResolvers<void>();
```
d3lm, line 10: Maybe newline before to separate let and const?

d3lm (commented):
Generally looks good. I only left minor comments. Great work!

A few additional things I noticed while testing this on my end.

I noticed is that there's a visual bug with the iframes when moving between the two sides which doesn't look great. Maybe not a lot of people will notice but it caught my eye immediately. Would be cool if we could investigate a way to fix this or at least make it less noticeable.

https://github.com/stackblitz/tutorialkit/assets/12571019/1a0a0370-16a5-44d0-8df0-15227f5d888f

Another thing is that there's the border visible on the left when switched to the right side. Visually that looks a bit off IMO.

CleanShot 2024-07-02 at 18 25 20

Also, we may need to think about the menu if we ship mobile support because right now it's not really usable as you can't read the titles. So we may need to go back to the design drawing board. Elements usually need to be adjusted to work well on mobile. The breadcrumb was mostly tailored for Desktop use. So I'd prefer if we could fix before shipping mobile support. I mean we can always fix it later but to me, if a user can't properly read the tiles then they can't properly navigate between different lessons.

image

Nemikolh:
> I noticed is that there's a visual bug with the iframes when moving between the two sides which doesn't look great. Maybe not a lot of people will notice but it caught my eye immediately. Would be cool if we could investigate a way to fix this or at least make it less noticeable.

I think that switching to `requestAnimationFrame` will make that bug disappear completely (it's the `setInterval` that's causing this because it does not update often enough). My worries was that it might cause a higher cpu usage but maybe it's fine.

> Another thing is that there's the border visible on the left when switched to the right side. Visually that looks a bit off IMO.

Oh good catch! :pray: 

> Also, we may need to think about the menu if we ship mobile support because right now it's not really usable as you can't read the titles. So we may need to go back to the design drawing board. Elements usually need to be adjusted to work well on mobile. The breadcrumb was mostly tailored for Desktop use. So I'd prefer if we could fix before shipping mobile support. I mean we can always fix it later but to me, if a user can't properly read the tiles then they can't properly navigate between different lessons.

Good point. What if we just show the lesson titles in the dropdown on mobile? 
Or maybe we split over two lines.

Nemikolh:
@d3lm I addressed all the comments in this PR and made simple changes. Namely:

 * The Breadcrumbs are now hidden and we only show the lesson title. We'll probably want to get some design there if we'd like to do  something different.
 * The tiny border on mobile on the left is now gone. :raised_hands: 
 * The bug that Sam reported is fixed as well. Turns out that when you do a build, `UnoCSS` does not sees the same content and the astro files must be added manually. This is not great at all, I'll try to create a minimal reproduction for UnoCSS to get some feedback on whether this behaviour is expected or not.
 * I removed the resize logic and we now use `requestAnimationFrame` to update the iframe positions. Works like a charm.

d3lm (approved):
Great work and thanks for addressing the feedback 🙏 This looks good 👏

PR #92, opened 2024-06-24
fix: support a base different from / in astro config

### LINKED ISSUE #89: Base deploy path

Currently unable to serve a tutorial with anything other than the default base path of `/`

Leverage the `import.meta.env.BASE_URL` property that can be configured in the `astro.config` to correctly prefix paths within generated tutorials.

N/A

### Additional context

I was able to track down most of the places that needed to be updated. There might be some more that I didn't encounter while I was working through the issues.

- **Logo.astro** : logo file path
- **Layout.astro** : favicon
- **index.astro** : favicon and redirect url
- **content.ts** : lesson next/prev href
- **nav.ts** : href
- **lesson-files.ts** : fetch

Nemikolh:
This PR makes it possible to use a base different from `/` in Astro configuration.

Closes #89


packages/astro/src/default/components/Logo.astro
```diff
@@ -16,7 +16,7 @@ function readLogoFile(logoPrefix: string) {
     const exists = fs.existsSync(path.join('public', logoFilename));
 
     if (exists) {
-      logo = `/${logoFilename}`;
+      logo = `${import.meta.env.BASE_URL}/${logoFilename}`;
```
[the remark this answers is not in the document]
Nemikolh: Oh good catch! I only tested with a base and assumed that this would be an empty string when no base is set. :see_no_evil:

AriPerkkio (approved):
Looks good! Here's deployment made with these changes: https://ariperkkio.github.io/tutorial-vite-plugin

packages/astro/src/default/utils/url.ts
```diff
@@ -0,0 +1,18 @@
+export function joinPaths(baseURL: string, ...paths: string[]): string {
```
SamVerschueren, line 1: Can't we use something like `URL()` to join the paths?

If not, I'd like to see some tests for this method 🙈 .

Nemikolh: I thought about that at first but wasn't sure it would work. Totally agree on adding tests. Will do that now :+1:

packages/runtime/src/lesson-files.ts
```diff
@@ -63,7 +65,7 @@ export class LessonFilesFetcher {
     this._templateLoadTask?.cancel();
 
     const task = newTask(async (signal) => {
-      const response = await fetch(`/${templatePathname}`, { signal });
+      const response = await fetch(`${this._baseURL}${templatePathname}`, { signal });
```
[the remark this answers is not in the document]
Nemikolh: Yep that's the assumption. I'll add a comment to the constructor about the assumption. Also I think I was wrong to call this `baseURL` it's more `basePathname` as it's not a URL.

I don't think we can use `new URL` here because of that.

packages/astro/src/default/utils/url.ts
Nemikolh: Oh nice after adding tests I found a bug (not a current bug in our code but one that could eventually happen in the future). It's when all the arguments are empty strings.

packages/runtime/src/lesson-files.ts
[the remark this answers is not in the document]
SamVerschueren: I don't like making assumptions tbqh, especially not with things like this. As I said, can't we use `new URL()` or `path.join()` or something?

packages/astro/src/default/utils/url.ts
SamVerschueren: In what scenario's wouldn't something like `new URL()` work? I'd rather use something like that then a custom implementation.

packages/runtime/src/lesson-files.ts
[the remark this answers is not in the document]
Nemikolh: No we can't because it's not an URL. We can't use `path` because this code is not in node. And the `joinPaths` utility is not available in this package.

packages/astro/src/default/utils/url.ts
Nemikolh: No it doesn't because baseURL is not a baseURL. Oh gosh I just realised that the first argument is named incorrectly. I'll fix that :+1:

packages/runtime/src/store/index.ts
```diff
@@ -14,7 +14,12 @@ import { TerminalStore } from './terminal.js';
 
 interface StoreOptions {
   webcontainer: Promise<WebContainer>;
+
+  /** Whether or not authentication is used for the WebContainer API */
```
SamVerschueren, line 18: I would make this a multiple line comment. Also, it should end with a `.`.

packages/astro/scripts/build.js
```diff
@@ -34,3 +35,6 @@ if (existsSync('./dist/default')) {
 
 // copy default folder unmodified
 await cp('./src/default', './dist/default', { recursive: true });
+
+// remove tests files
```
d3lm, line 39:
```suggestion
// remove test files
```

d3lm: Also, can't we just exclude the `.spec.ts` files from being emitted to the `dist` folder?

Nemikolh: The default folder is copied verbatim on line 37. This is why we have to remove those files after copying them.

PR #93, opened 2024-06-24
feat: allow custom `src/pages/index.astro`

Nemikolh:
This PR makes it possible to have a custom `src/pages/index.astro` by modifying the `defaultRoutes` to support an extra mode in which it does not inject the default route for `/{base}`.

Once this PR is merged, to enable this feature on a project, one would have to do:

```ts

export default defineConfig({
  integrations: [
    tutorialkit({
      // only enable the routes for the tutorial
      defaultRoutes: 'tutorial-only',
    }),
  ],
});

```

The alternative to this I considered was to modify `index.astro` so that it reads file from the `src/pages/index.astro` folder, however this wouldn't have worked under the `eject` mode as the file would end up reading itself recursively.


packages/astro/src/index.ts
```diff
@@ -15,9 +15,9 @@ export interface Options {
    *
    * Set this to false to customize the pages.
    *
-   * @default true
+   * @default 'all'
```
Nemikolh, line 18:
```suggestion
   * @default true
```

d3lm (approved):

d3lm:
When adding these things we need to make sure to add this to the docs as well and it'd be great if we can link the docs PR here as well for full transparency.

SamVerschueren:
> it'd be great if we can link the docs PR here as well for full transparency.

It'd be better to just update the docs in this PR. So it's a single PR which includes everything.

d3lm:
Right! Forgot for a second that the docs are in the same repo 🤦‍♂️

Nemikolh:
@d3lm, @SamVerschueren I've updated the JSDoc of the property so that the new value is explained.

Given we currently don't have any docs for our JSDocs I don't think it makes sense that I introduce that as part of this PR?

What do you think?

PR #95, opened 2024-06-25
chore: remove pnpm from tool versions file

d3lm:
This PR removes `pnpm` from the `.tool-versions` file because it's unfortunately currently unsupported by Cloudflare Pages. Instead I added `pnpm` to the `engines` field to enforce a specific version. That version is also enforced in the build options for Cloudflare Pages.


Nemikolh (approved):
Nice!! :star_struck:

.github/workflows/ci.yaml
```diff
@@ -25,7 +25,7 @@ jobs:
       - name: Setup
         uses: pnpm/action-setup@v4
         with:
-          version: 8
+          version: 8.15.6
```
d3lm, line 28: Using this version because it's currently the latest we support on StackBlitz.com.

PR #99, opened 2024-06-26
feat: template inheritance

Nemikolh:
This PR makes it possible to extend a template with another template by adding a configuration file, `.tk-config.json` with the following content:

```json
{
  "extends": "../path-to-other-template"
}
```

So with the following structure:

```
templates
├── vite-app
│   ├── index.html
│   ├── main.js
│   ├── package.json
│   ├── package-lock.json
│   └── style.css
└── vite-app-2
    ├── index.html
    └── .tk-config.json
```

The content of `vite-app-2` and `vite-app` will be identical except for the `index.html` which will be the one living in `vite-app`.

### Notes

* Because this is a path, it's also possible to do something like `../../node_modules/tuto-templates/foo`. So this means that if you have fixed templates that lives somewhere else, you can extend them. They won't be watched by the watcher though, so if they change it won't be reflected.
* This PR technically  also adds support for `_files` and `_solution` because it made the code simpler. It might be something people want to use in the future though so this is probably a good thing?
* :point_right: This PR also refactored a bit the `webcontainer-files.ts` a bit with extra tests to make it easier to maintain this code.


packages/astro/src/webcontainer-files/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const IGNORED_FILES = ['**/.DS_Store', '**/*.swp'];
+export const EXTEND_CONFIG_FILEPATH = '/.tk-config.json';
```
Nemikolh, line 2: Interested to hear feedback / thoughts on the name of the config file here. :eyes: 

I started with `template.json`, then renamed it to `.template.json`, then `.config.json` to finally going with `.tk-config.json`.

My reasoning has been: I started with `template.json` because this is for templates. Then I thought, this might collide with user files, so added the starting `.`. I then realised that the code would also work for `_files` and `_solution` because we don't distinguish them and so thought maybe a more generic name would be better. I went with `.config.json` but figured that adding a `tk-` prefix would make it more explicitely TutorialKit specific and help tooling like the VSCode extension, to provide a schema for the JSON.

cc @sulco

packages/astro/src/webcontainer-files/filesmap.ts
```diff
@@ -0,0 +1,218 @@
...
+    if (!node) {
+      node = new FilesMap(resolvedPath);
+      this._nodes.set(resolvedPath, node);
+    }
+
+    node.update(this._nodes, logger);
+  }
+}
+
+export class FilesMap {
```
Nemikolh, line 46: I created this class to make it easier to understand the graph structure and encapsulate the logic around:
* Cycle detection (which should not throw or cause an infinite loop)
* The conversion from a node in that graph to a `Files` entity that can be JSON serialized later

packages/astro/src/webcontainer-files/filesmap.ts
```diff
@@ -0,0 +1,218 @@
...
+  }
+}
+
+function webcontainerPath(dir: string, filePath: string) {
+  const result = `/${path.relative(dir, filePath)}`;
+
+  // mostly for Windows
+  if (path.sep !== '/') {
+    return result.replaceAll('\\', '/');
+  }
```
Nemikolh, line 215: I think this should fix a bug on windows which I think the current TutorialKit has where files inside sub-folders will use `\\` when they should use `/`.

AriPerkkio (commented):
I've now tested this locally and it works really well, nice! 🙌 

Here you can see how much code repetition I could reduce in a project that has 4 templates: https://github.com/AriPerkkio/tutorial-vite-plugin/pull/6

```diff
src/templates
  ├── default
  │   ├── content.yaml
  │   ├── index.html
  │   ├── index.js
  │   ├── main.js
  │   ├── package-lock.json
  │   ├── package.json
  │   ├── public
  │   │   └── vite.svg
  │   ├── style.css
  │   ├── vite.config.internal.ts
  │   └── vite.config.ts
  │
  ├── empty  # Template that's used in terminal-only Node lesson
  │   └── index.js
  │
  ├── env-plugin
+ │   ├── .tk-config.json   # { "extends": "../default" }
- │   ├── index.html
  │   ├── index.js
- │   ├── main.js
- │   ├── package-lock.json
- │   ├── package.json
- │   ├── public
- │   │   └── vite.svg
  │   ├── style.css
- │   ├── vite.config.internal.ts
- │   └── vite.config.ts
  │
  └── replace-plugin
+     ├── .tk-config.json   # { "extends": "../default" }
-     ├── index.html
      ├── main.js
-     ├── package-lock.json
-     ├── package.json
-     ├── public
-     │   └── vite.svg
      ├── style.css
      ├── tutorial-example.js
-     ├── vite.config.internal.ts
-     └── vite.config.ts
```

packages/astro/src/webcontainer-files/cache.ts
```diff
@@ -0,0 +1,198 @@
...
+
+    while (this._requestsQueue.size > 0) {
+      // first we update all files map
+      for (const folderPath of this._requestsQueue) {
+        this._filesMapGraph.updateFilesMapByFolder(folderPath, this._logger);
+      }
+
+      // then we make sure all their dependencies are added as well
+      for (const folderPath of this._requestsQueue) {
+        for (const dependency of this._filesMapGraph.getFilesMapByFolder(folderPath)!.allDependents()) {
```
AriPerkkio, line 125: Should the falsy return values of `getFilesMapByFolder` be handled here instead? Or make `getFilesMapByFolder` return an object with dummy `.allDependents()` instead of `undefined`?

Nemikolh:
@AriPerkkio That is indeed a lot!! :exploding_head: :exploding_head: 

Your PR alone really show how useful this feature is! Thanks for suggesting it :raised_hands:

packages/astro/src/webcontainer-files/filesmap.spec.ts
```diff
@@ -0,0 +1,271 @@
...
+      {
+        "/file1": "",
+      }
+    `);
+    expect(await node2?.toFiles(logger as any)).toMatchInlineSnapshot(`
+      {
+        "/file2": "",
+      }
+    `);
+  });
```
AriPerkkio, line 47: Tip: `expect()` doesn't do type narrowing but `assert()` does: https://vitest.dev/api/assert.html#assert-1

```suggestion
    assert(node1 !== undefined)
    assert(node2 !== undefined)
    expect(logger.warn).not.toHaveBeenCalled();

    expect(await node1.toFiles(logger as any)).toMatchInlineSnapshot(`
      {
        "/file1": "",
      }
    `);
    expect(await node2.toFiles(logger as any)).toMatchInlineSnapshot(`
      {
        "/file2": "",
      }
    `);
  });
```

packages/astro/src/webcontainer-files/filesmap.ts
AriPerkkio, line 215: I guess `path.sep` can only be `/` or `\` but following would handle all possible edge cases:

```suggestion
  // normalize path separators
  if (path.sep !== '/') {
    return result.replaceAll(path.sep, '/');
  }
```

packages/template/src/templates/vite-app-2/.tk-config.json
AriPerkkio, line 1: Would be good to use this `vite-app-2` in some lesson in `packages/template/content`.

packages/astro/src/webcontainer-files/cache.ts
Nemikolh: It cannot return a falsy value right now because `updateFilesMapByFolder` called in the previous loop always adds the folder it received to the graph.

It could return a falsy value if `updateFilesMapByFolder` was modified in the future and would no longer provide that guarantee.

What about logging a `This is a bug!` message if the value is `undefined`? Wdyt?

docs/tutorialkit.dev/src/content/docs/guides/creating-content.mdx
```diff
@@ -109,3 +109,11 @@ template: my-advanced-template
 ```
 
 This declaration will make TutorialKit use the `src/templates/my-advanced-template` directory as the base for the lesson.
+
+If you start having a lot of templates and they all share some code, you can create a shared template that they all extend. This way, you can keep the shared code in one place and avoid duplication. To do that, you need to specify the `extends` property in the template's `.tk-config.json` file:
```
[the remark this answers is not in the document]
Nemikolh: Oh that's a good idea! Will add it :+1:

packages/astro/src/webcontainer-files/filesmap.spec.ts
Nemikolh: Oh that's super nice!! I'll update all the other tests

packages/astro/src/webcontainer-files/filesmap.ts
Nemikolh: Good point!

packages/astro/src/webcontainer-files/cache.ts
AriPerkkio: Oh right, it's just TS that doesn't necessary keep up with the types here. I think this is fine as is. In case of bug it will still throw on `.allDependents` access.

AriPerkkio (approved):

PR #101, opened 2024-06-27
ci: detect `pnpm` version from `package.json`

AriPerkkio:
The `pnpm/action-setup@v4` can detect `pnpm` version from `package.json`. In order to have `package.json` present on file system, we need to run `actions/checkout@v4` first.


Nemikolh (approved):
Nice improvement! :star_struck:

PR #103, opened 2024-06-28
docs: move configuration to reference section

SamVerschueren:
This PR moves the `Configuration` page to the `Reference` section as previously discussed.


docs/tutorialkit.dev/astro.config.ts
```diff
@@ -19,11 +19,7 @@ export default defineConfig({
         light: './src/assets/brand/tutorialkit-logo-light.svg',
         replacesTitle: true,
       },
-      plugins: [
-        starlightLinksValidator({
-          exclude: ['../../guides/**/*'],
-        }),
-      ],
+      plugins: [starlightLinksValidator()],
```
SamVerschueren, line 22: By using absolute paths instead, this seems to work. I only just noticed that other pages also use absolute paths 🤷 .

Nemikolh (approved):
Looks good! Thanks a lot for the PR! :smiley:

PR #104, opened 2024-06-28
feat: make core react components easily accessible

Nemikolh:
This PR changes a bit the `@tutorialkit/react-components` so that the `CodeMirrorEditor`, `Terminal` and `FileTree` can be easily accessed under a `/core` module:

```ts
import { Terminal, CodeMirrorEditor, FileTree } from '@tutorialkit/components-react/core';
```

Given that the `Terminal` component is not SSR friendly because of `@xterm`, it is also possible to import the component individually, via `/core/<component name>` and both default and named exports are provided:

```ts
import Terminal from '@tutorialkit/components-react/core/Terminal';
import CodeMirrorEditor from '@tutorialkit/components-react/core/CodeMirrorEditor';
import FileTree from '@tutorialkit/components-react/core/FileTree';
```

(in the astro integration we lazy import the terminal with a dynamic import and add a `<Suspense>` boundary to make it work)


packages/components/react/src/core/CodeMirrorEditor/index.tsx
```diff
@@ -176,13 +178,15 @@ export function CodeMirrorEditor({
   }, [doc]);
 
   return (
-    <div className="h-full relative">
```
Nemikolh, line 179: I've moved the responsability of setting the height to the caller. While working on an example where we wanted a fix height, this felt nicer than to wrap it in a div.

However, maybe in the majority of use cases, `h-full` is preferred. So maybe we should default `className` to `h-full`? I'm not sure.

packages/components/react/src/index.ts
```diff
@@ -1,10 +1,8 @@
 export * from './BootScreen.js';
-export * from './CodeMirrorEditor/index.js';
-export * from './FileTree.js';
```
Nemikolh, line 3: Removed those exports because they're not used at the moment and the intent is to have them imported via the `/core` module.

d3lm (approved):
Nice one!

packages/components/react/src/core.ts
```diff
@@ -0,0 +1,3 @@
+export * from './core/CodeMirrorEditor/index.js';
+export * from './core/FileTree.js';
+export * from './core/Terminal/index.js';
```
AriPerkkio, line 3: As this now becomes part of the library's public API, I would recommend to be explicit about the exports here. It's better to be verbose with these exports so that we'll know which parts were "promised" to be present.

```suggestion
export { CodeMirrorEditor } from './core/CodeMirrorEditor/index.js'; // and maybe the types too if they were intended?
export { FileTree } from './core/FileTree.js';
export { Terminal, type TerminalRef } from './core/Terminal/index.js';

```

packages/components/react/src/core/CodeMirrorEditor/index.tsx
```diff
@@ -176,13 +178,15 @@ export function CodeMirrorEditor({
   }, [doc?.value, doc?.filePath, doc?.loading]);
 
   return (
-    <div className="h-full relative">
+    <div className={`relative ${className}`}>
```
AriPerkkio, line 181: I think we have `classnames` package available in the project:

```suggestion
    <div className={classNames('relative', className)}>
```

AriPerkkio: Same for Terminal.

packages/components/react/src/core.ts
AriPerkkio: Oh well, `packages/components/react/src/index.ts` is also exporting quite much using wildcards 🤔 
Feel free to leave this as is, but at some point I think we'll need to check which entrypoints are exposed to users. Right know this is quite prone to accidental breaking changes.

packages/components/react/package.json
```diff
@@ -8,11 +8,31 @@
...
+        "dist/core/CodeMirrorEditor/index.d.ts"
+      ],
+      "./core/FileTree": [
+        "dist/core/FileTree.d.ts"
+      ],
+      "./core/Terminal": [
+        "dist/core/Terminal/index.d.ts"
+      ]
+    }
+  },
```
AriPerkkio, line 35: We could declare these in the `exports` directly, e.g.

```json
  "exports": {
    ".": {
      "import": {
        "types": "./dist/index.d.ts",
        "default": "./dist/index.js"
      }
    },
    "./core": {
      "import": {
        "types": "./dist/core.d.ts",
        "default": "./dist/core.js"
      }
    },
    "./core/CodeMirrorEditor": {
      "types": "./dist/core/CodeMirrorEditor/index.d.ts",
      "default": "./dist/core/CodeMirrorEditor/index.js"
    },
```

AriPerkkio (commented):
Some thoughts but otherwise it looks good to me.

packages/components/react/src/core.ts
Nemikolh: That's a very good point! I'll add explicit imports for this file, we can revisit the other imports in separate PRs.

packages/components/react/src/core/CodeMirrorEditor/index.tsx
Nemikolh: We do! Does that make sense though to use it here given a simple string concatenation works and is more efficient?

Now that said, happy to change it if you feel strongly about it.

packages/components/react/package.json
Nemikolh: Oh this is nice! I'll give it a go :raised_hands:

packages/components/react/src/core/CodeMirrorEditor/index.tsx
AriPerkkio: I've usually used `classNames` for all conditional or more complex `className` declarations. It makes sure no `undefined` or `false` classnames end up in the final markdown. And keeps stuff consistent. 

Here the `className` prop is given default value of `''`, which may end up as `null|false` if end user passes falsy value as prop, e.g. `<CodeMirrorEditor className={condition && "mt-2"} />`. Using `<CodeMirrorEditor className={condition ? "mt-2" : undefined} />` works but shorthands are always nice. 😄

Nemikolh: Gotcha! I'm convinced :+1:

AriPerkkio (approved):

PR #106, opened 2024-06-28
ci: release workflow

AriPerkkio:
Adds scripts for performing releases.

- Start by running `Prepare release PR` workflow:

- This workflow will bump versions and generate changelogs for all packages, except CLI. It creates a PR like https://github.com/AriPerkkio/tutorialkit/pull/26

- Merging the PR triggers another workflow with two jobs:
  - First one creates git tags and publishes packages to npm
  - It triggers second job which prepares release PR for CLI package, with the updated package versions: https://github.com/AriPerkkio/tutorialkit/pull/29

- The opened PR will run CLI Integration tests (this doesn't trigger automatically on my fork though, let's re-check this on first real release 🤔)
- Merging this PR will trigger another workflow that updates versions in templates, runs CLI integration tests (once again just to be sure) and finally publishes NPM package


.github/workflows/publish-release.yaml
```diff
@@ -0,0 +1,111 @@
...
+    branches:
+      - main
+
+jobs:
+  publish_release:
+    name: Publish release
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      id-token: write
```
AriPerkkio, line 14: These are required for `--provenance` flag.

.github/workflows/publish-release.yaml
```diff
@@ -0,0 +1,111 @@
...
+        id: resolve-release-version
+
+      - name: Publish to npm
+        run: >
+          pnpm --recursive
+          --filter "@tutorialkit/*"
+          --filter create-tutorial
+          exec pnpm publish --provenance
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
AriPerkkio, line 36: We'll need to manually add this secret to the project.

.github/actions/setup-and-build/action.yml
```diff
@@ -0,0 +1,25 @@
+name: Setup and build
```
AriPerkkio, line 1: This is a local Github action that can be used to reduce repetition. See https://github.com/stackblitz/tutorialkit/pull/106/commits/cddfde9deeb9ca01d7b05c894eceaba16c97e354 for the affected files.

AriPerkkio:
Need to recheck the integration tests again. They started to fail for me locally after rebase. Would appriciate if someone else could also try them locally. 🙌

.github/actions/resolve-release-version/action.yml
```diff
@@ -0,0 +1,24 @@
...
+      shell: bash
+      run: >
+        echo "$(
+          node -e "
+            const fs = require('fs');
+            const pkg = fs.readFileSync('./packages/astro/package.json', 'utf8');
+            const version = JSON.parse(pkg).version
+            console.log('version=' + version);
+          "
+        )" >> $GITHUB_OUTPUT
```
Nemikolh, line 24: Oh this is nice! I feel llike we should use `jq` though. 

```suggestion
        echo "version=$(jq -r .version ./packages/astro/package.json)" >> $GITHUB_OUTPUT
```

(only because we do that in `webcontainer`, if you prefer the node version we can keep it)

.github/workflows/prepare-release.yaml
```diff
@@ -0,0 +1,44 @@
...
+        run: >
+          pnpm --recursive
+          --filter "@tutorialkit/*"
+          --filter create-tutorial
+          exec pnpm version --no-git-tag-version --allow-same-version ${{ inputs.version }}
+
+      - name: Generage changelog
+        run: |
+          pnpm run changelog
+          git checkout ./packages/cli
```
Nemikolh, line 34: I think that we should keep the CLI changelog here. 

IIRC once we set the tag, the changelog will use the latest tag known in git and use that to update the changelog. So that means that then it will generate an empty changelog for the CLI.

But I might be wrong.

Have you verified that doing it later works?

.github/workflows/publish-release.yaml
```diff
@@ -0,0 +1,111 @@
...
+          pnpm --recursive
+          --filter tutorialkit
+          exec npm version --no-git-tag-version --allow-same-version ${{ steps.resolve-release-version.outputs.version }}
+
+      - name: Generage changelog
+        run: |
+          pnpm run changelog
+          git add ./packages/cli ./packages/template
+          git checkout ./packages
+          git checkout CHANGELOG.md
```
Nemikolh, line 72: So this is the step I would remove as my guess is that I think it won't work because the tag for the latest version was already pushed.

Nemikolh (commented):
Looks really good! :muscle:

.github/workflows/prepare-release.yaml
AriPerkkio: Changelog for CLI is indeed empty after these steps. Let's keep the CLI's changelog here. 👍

.github/workflows/prepare-release.yaml
```diff
@@ -0,0 +1,44 @@
...
+        with:
+          fetch-depth: 0
+
+      - uses: ./.github/actions/setup-and-build
+
+      - name: Bump versions
+        run: >
+          pnpm --recursive
+          --filter "@tutorialkit/*"
+          --filter create-tutorial
```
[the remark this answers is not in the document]
AriPerkkio: Oh good catch! So it's the `@tutorialkit/*` packages that are used by `tutorialkit` and `create-tutorial` packages. 
I'll change the first release phase to consider only `@tutorialkit/*` packages, and the second one to handle `tutorialkit` and `create-tutorialkit`.

.github/actions/resolve-release-version/action.yml
AriPerkkio: @Nemikolh is `jq` available on CI without any additional setups? Happy to switch into that one.

packages/cli/src/commands/create/index.ts
```diff
@@ -253,30 +251,26 @@ function updatePackageJson(dest: string, projectName: string, flags: CreateOptio
   }
 
   const pkgPath = path.resolve(dest, 'package.json');
-
   const pkgJson = JSON.parse(fs.readFileSync(pkgPath, 'utf8'));
 
   pkgJson.name = projectName;
 
-  updateWorkspaceVersions(pkgJson.dependencies, TUTORIALKIT_VERSION);
-  updateWorkspaceVersions(pkgJson.devDependencies, TUTORIALKIT_VERSION);
```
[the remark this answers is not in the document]
AriPerkkio: I'm not yet familiar with these CLI commands. This change came from https://github.com/stackblitz/tutorialkit/commit/e444b43035711de4128ce8de79ee3ecb6b5d3294#diff-e85db3b9733aa67073e3397b203523dcd174b6a7b2ee32e2ef6e53933e9b9451.

So the issue here now is that when running `node ./packages/cli/dist/index.js create`, the created tutorial has workspace-protocol in `@tutorialkit` dependencies? It should be using the actual version numbers there?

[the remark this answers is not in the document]
Nemikolh: Yes, that's indeed the issue! :smiley: 

Alternatively we could remove the symlink and instead have the build step always copy the template & apply the override. I would be totally fine with that change.

.github/workflows/publish-release.yaml
Nemikolh: @d3lm I think if you can add @stackblitz-gitops to the packages, then we should be able to reuse the token of @stackblitz-gitops.

@apai4 Do you think that would be fine? Or should we have a different token?

.github/workflows/prepare-release.yaml
```diff
@@ -0,0 +1,41 @@
+name: Prepare release PR
```
d3lm, line 1:
```suggestion
name: Prepare Release PR
```

.github/workflows/integration-test-cli.yaml
```diff
@@ -0,0 +1,27 @@
+name: Integration Tests CLI
```
d3lm, line 1: Should this be the same as the name of the job or do we want this to be different?
```suggestion
name: CLI Integration Tests
```

.github/workflows/publish-release.yaml
```diff
@@ -0,0 +1,108 @@
+name: Publish release
```
d3lm, line 1:
```suggestion
name: Publish Release
```

.github/workflows/publish-release.yaml
```diff
@@ -0,0 +1,108 @@
...
+    if: ${{ contains(github.event.head_commit.message, 'release @tutorialkit packages') }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: ./.github/actions/setup-and-build
+
+      # Sets steps.resolve-release-version.outputs.version
```
d3lm, line 25:
```suggestion
      # sets steps.resolve-release-version.outputs.version
```

.github/workflows/publish-release.yaml
```diff
@@ -0,0 +1,108 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: ./.github/actions/setup-and-build
+
+      # Sets steps.resolve-release-version.outputs.version
```
d3lm, line 56:
```suggestion
      # sets steps.resolve-release-version.outputs.version
```

.gitignore
```diff
@@ -18,7 +18,7 @@ dist-ssr
 *.sln
 *.sw?
 .pnpm-store
-/tutorialkit/template
+/packages/cli/template
```
d3lm, line 21: Good catch!

integration/cli/create-tutorial.test.ts
d3lm, line 1: Are these new package manager tests supposed to use a lock file? Maybe I missed it but I didn't find them in the snapshot. For example I expected to find a `yarn.lock` when `yarn` was used. Did I miss that somewhere?

packages/cli/scripts/update-template.js
```diff
@@ -0,0 +1,100 @@
...
+  await execa('pnpm', ['install', '--lockfile-only'], { cwd: tmp, env });
+  await execa('yarn', ['install'], { cwd: tmp, env });
+
+  fs.cpSync(path.join(tmp, 'package-lock.json'), path.join(templateDest, 'package-lock.json'));
+  fs.cpSync(path.join(tmp, 'pnpm-lock.yaml'), path.join(templateDest, 'pnpm-lock.yaml'));
+  fs.cpSync(path.join(tmp, 'yarn.lock'), path.join(templateDest, 'yarn.lock'));
+
+  console.log('Created', path.join(templateDest, 'package-lock.json'));
+  console.log('Created', path.join(templateDest, 'pnpm-lock.yaml'));
+  console.log('Created', path.join(templateDest, 'yarn.lock'));
```
d3lm, line 78: Maybe we can use `success` from the logger module for these logs to indicate more visually that it successful created the files.

d3lm, line 78:
```suggestion
  success(`Created ${path.join(templateDest, 'package-lock.json')}`);
  success(`Created ${path.join(templateDest, 'pnpm-lock.yaml')}`);
  success(`Created ${path.join(templateDest, 'yarn.lock')}`);
```

packages/cli/src/commands/create/index.ts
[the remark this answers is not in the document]
d3lm: I would personally prefer the symlink so we don't copy so many files during dev.

d3lm (commented):
Left a few minor comments. Great work @AriPerkkio 👏

.github/workflows/publish-release.yaml
d3lm: I pinged Pai about this because I cannot add NPM access tokens since I don't have permissions to generate them.

apai4: I'll create a new token for `stackblitz-devops` on NPM for this repo and add it to `secrets.NPM_TOKEN`. `stackblitz-devops` will need write access to the NPM package.

apai4: The token has been created and added to the repo. Let me know if it works once `stackblitz-gitops` is granted access

Nemikolh: @d3lm can you grant access to `stackblitz-gitops` on all the packages?

packages/cli/src/commands/create/index.ts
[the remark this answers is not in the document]
Nemikolh: Note that in dev, it would only happen when building on the CLI but not when doing `pnpm template:dev`.
 
Right now you don't see the impact of the overrides on dev unless you ran `prepack` instead of `build`. If we remove the symlink and always do the copy it would. 

I think it would also make the CLI a bit simpler.

With all that said I'm fine keeping the symlink.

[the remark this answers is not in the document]
d3lm: Yea true. I mean as long as dev is not impacted much and we can keep the symlink there then that sounds good to me. Just wanna make sure the DX is still smooth and we avoid file copies if possible.

.github/workflows/integration-test-cli.yaml
AriPerkkio: We can use the same name, I'll apply this suggestion.

integration/cli/create-tutorial.test.ts
AriPerkkio: I'll need to re-check the test files 👀 . I've just applied the changes from https://github.com/stackblitz/tutorialkit/commit/e444b43035711de4128ce8de79ee3ecb6b5d3294.

AriPerkkio: Lockfiles are added to `/<root>/tutorialkit/packages/cli/template`, but create command copies template from `/<root>/tutorialkit/packages/cli/_template`.

@Nemikolh what's the difference between these two directories? Should we use only one of them?

Nemikolh: `_template` is the symlink used when you only do a `build` on the CLI (so typically what you would use by default in dev) and `template` is the folder that gets created when doing a `prepack`.

You can even use a different folder for the template in dev by setting the `TUTORIALKIT_TEMPLATE_PATH` environment variable and then do a `pnpm --filter=tutorialkit build`

packages/cli/src/commands/create/dependencies.ts
```diff
@@ -0,0 +1,127 @@
...
+
+const LOCK_FILES = new Map<PackageManager, string>([
+  ['npm', 'package-lock.json'],
+  ['pnpm', 'pnpm-lock.yaml'],
+  ['yarn', 'yarn.lock'],
+]);
+
+export type PackageManager = 'npm' | 'yarn' | 'pnpm' | 'bun';
+
+export async function installDependencies(cwd: string, flags: CreateOptions) {
```
AriPerkkio, line 19: This file is actually not used. It's mostly doing the same as `packages/cli/src/commands/create/package-manager.ts`. What was the intention of this @SamVerschueren?

integration/cli/create-tutorial.test.ts
AriPerkkio: Lockfiles are now present in the snapshots. I've used separate snapshot files for each package manager to make things more visible.

Also I had to exclude the `should create and build a project` until `@tutorialkit/theme` has been released.

packages/cli/src/commands/create/dependencies.ts
SamVerschueren: This is probably coming from my branch?

I created my branch before I did the refactor in https://github.com/stackblitz/tutorialkit/pull/77 where I removed this file and created `package-manager.ts`. So we should be good by removing this entirely.

AriPerkkio: Thanks, that makes sense! I'll need to apply these changes to `package-manager.ts` instead and remove this old `dependencies.ts`.

AriPerkkio: These chages are now in visible in https://github.com/stackblitz/tutorialkit/pull/106/commits/32fc233c188d2424a83fd992c57dbb8cca66f3c6. From snapshots it's visible how unrelated lock files are removed when package manager is selected.

packages/cli/src/commands/create/package-manager.ts
```diff
@@ -6,7 +8,27 @@ import { DEFAULT_VALUES, type CreateOptions } from './options.js';
...
+  ['yarn', 'yarn.lock'],
+]);
+
+export async function selectPackageManager(cwd: string, flags: CreateOptions) {
+  const packageManager = await resolvePackageManager(flags);
+
+  // remove lock files for other package managers
+  for (const [pkgManager, lockFile] of LOCK_FILES) {
+    if (pkgManager !== packageManager) {
+      console.log('Removing', path.join(cwd, lockFile));
```
Nemikolh, line 23: Oh this `console.log` should probably be removed. Or if we want to keep it then we should use `prompts.log.info`.

AriPerkkio: Ugh, my own debug logs. Good catch.

packages/cli/package.json
```diff
@@ -16,9 +16,7 @@
   },
   "scripts": {
     "build": "node scripts/build.js",
-    "prepack": "node scripts/pre-pack.js",
-    "postpack": "node scripts/cleanup.js",
-    "test": "vitest --testTimeout=300000"
```
Nemikolh, line 21: Is it intentional to have the `test` script removed? I think we should keep it to keep testing the tests in the CLI package.

AriPerkkio: This is also coming from https://github.com/stackblitz/tutorialkit/commit/e444b43035711de4128ce8de79ee3ecb6b5d3294#diff-6e2e2a1851648938b325ba84de634407a4e69a644ea61102df15ca4a8a7a9758 🤔 

Looking at the `packages/cli/test`, I think the `<root>/integration` is intended to replace these. But there's test like `cannot create project without installing but with starting` that we should still keep running from here. I'll add `test` script back, `vitest` dependency back and remove those duplicate test cases.

Nemikolh: I think it might indeed have been the case. However, there's value in keeping them because:
* They tests continuously the CLI with the workspace version
* Integrations are intended to test with the released version

However, I'm not 100% sure what @SamVerschueren had in mind. Sam am I correct? Or did you intend for those to be removed?

AriPerkkio: Ok, let's keep both. 👍

AriPerkkio:
Verified that everything still works after PR changes in my fork ✅ 

- https://github.com/AriPerkkio/tutorialkit/actions/runs/9762884812
- https://github.com/AriPerkkio/tutorialkit/pull/34
- https://github.com/AriPerkkio/tutorialkit/actions/runs/9762921311
- https://github.com/AriPerkkio/tutorialkit/pull/35
- https://github.com/AriPerkkio/tutorialkit/actions/runs/9762953469

.github/workflows/publish-release.yaml
d3lm: Yes, I can do that!

packages/cli/scripts/update-template.js
Nemikolh, line 1: Maybe this should be called `release-build.js`? Because it run the `build.js` script in "prod mode" and do the extra steps to do a release build.

Wdyt?

AriPerkkio: I agree. But does `release-build.js` sound like "Release the build"? Maybe instead use `build-release.js`, so that it's like "Build the release" - not actually releasing?

Nemikolh: Oh yeah good point!

.github/workflows/publish-release.yaml
d3lm: Invite is out. I think @apai4 needs to accept it.

Nemikolh (approved):
Fantastic work on this @AriPerkkio ! :smiley: :clap:

.github/workflows/prepare-release.yaml
```diff
@@ -0,0 +1,41 @@
...
+    inputs:
+      version:
+        description: 'Version to publish, e.g. 0.0.1'
+        required: true
+        default: '0.0.1'
+        type: string
+
+jobs:
+  prepare_release:
+    name: Prepare release PR
```
d3lm, line 14: Nit: Maybe we use the same name as for the workflow? Just for consistency.
```suggestion
    name: Prepare Release PR
```

.github/workflows/publish-release.yaml
```diff
@@ -0,0 +1,108 @@
+name: Publish Release
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  publish_release:
+    name: Publish release
```
d3lm, line 10: Same here?
```suggestion
    name: Publish Release
```

.github/workflows/publish-release.yaml
```diff
@@ -0,0 +1,108 @@
...
+
+      - name: Create and push git tag
+        run: |
+          git config --global user.name 'github-actions'
+          git config --global user.email 'github-actions@users.noreply.github.com'
+          git tag v${{ steps.resolve-release-version.outputs.version }}
+          git push origin v${{ steps.resolve-release-version.outputs.version }}
+
+  prepare_cli_release:
+    name: Prepare release for CLI
```
d3lm, line 45:
```suggestion
    name: Prepare Release for CLI
```

.github/workflows/publish-release.yaml
```diff
@@ -0,0 +1,108 @@
...
+        with:
+          # Note: `publish-release.yaml` checks explicitly for this commit message
+          commit-message: 'chore: release tutorialkit CLI, version: ${{ steps.resolve-release-version.outputs.version }}'
+          title: 'chore: release tutorialkit CLI, version: ${{ steps.resolve-release-version.outputs.version }}'
+          body: 'Bump tutorialkit CLI to version ${{ steps.resolve-release-version.outputs.version }}'
+          reviewers: SamVerschueren,d3lm,Nemikolh,AriPerkkio
+          branch: chore/release-cli-${{ steps.resolve-release-version.outputs.version }}
+
+  publish_release_CLI:
+    name: Publish release CLI
```
d3lm, line 78:
```suggestion
    name: Publish Release CLI
```

integration/cli/__snapshots__/npm-created.json
```diff
@@ -0,0 +1,54 @@
...
+  "src/templates/default/javascript.svg",
+  "src/templates/default/main.js",
+  "src/templates/default/package-lock.json",
+  "src/templates/default/package.json",
+  "src/templates/default/public",
+  "src/templates/default/public/vite.svg",
+  "src/templates/default/style.css",
+  "tsconfig.json",
+  "uno.config.ts"
+]
```
d3lm, line 54: Newline at the end of this file missing.

integration/cli/__snapshots__/pnpm-created.json
```diff
@@ -0,0 +1,54 @@
...
+  "src/templates/default/javascript.svg",
+  "src/templates/default/main.js",
+  "src/templates/default/package-lock.json",
+  "src/templates/default/package.json",
+  "src/templates/default/public",
+  "src/templates/default/public/vite.svg",
+  "src/templates/default/style.css",
+  "tsconfig.json",
+  "uno.config.ts"
+]
```
d3lm, line 54: Newline missing.

integration/cli/__snapshots__/yarn-created.json
```diff
@@ -0,0 +1,54 @@
...
+  "src/templates/default/main.js",
+  "src/templates/default/package-lock.json",
+  "src/templates/default/package.json",
+  "src/templates/default/public",
+  "src/templates/default/public/vite.svg",
+  "src/templates/default/style.css",
+  "tsconfig.json",
+  "uno.config.ts",
+  "yarn.lock"
+]
```
d3lm, line 54: Newline missing.

d3lm (commented):
Awesome work! Only minor remarks.

integration/cli/__snapshots__/yarn-created.json
SamVerschueren: This is written by the snapshot itself. Unless we can modify the generator to write this line, I would be strongly against manually updating it every single time.

d3lm: Ah yea that's true. Changing automatically generated files isn't ideal so totally agree. Kind of annoying that it doesn't add the newline there.

integration/cli/__snapshots__/npm-created.json
AriPerkkio: File snapshots don't support this out-of-the-box. https://vitest.dev/guide/snapshot.html#file-snapshots

I'll need to check Vitest's sources if there's a way to add this manually. Maybe custom snapshot serializer 🤔

d3lm: Yea, if it's way too much work then I guess it's fine. Just caught my eye but I understand that this is a snapshot and automatically generated so maybe it's fine.

AriPerkkio: Oh actually we can add the newline manually. Snapshot comparison ignores it automatically.

integration/cli/__snapshots__/yarn-created.json
SamVerschueren: Yeah I agree, don't like it either.

integration/cli/__snapshots__/npm-created.json
d3lm: Ohhh that is awesome!

d3lm (approved):
Great work 💪 👏 🔥

.github/workflows/publish-release.yaml
AriPerkkio: @d3lm do you have permissions to check if this is completed now? Publish step failed at https://github.com/stackblitz/tutorialkit/actions/runs/9791481416/job/27035364611

Issue #107, opened 2024-06-28
Problem with creating lessons without solution files


leonyoung1:
### Describe the bug

When there is a lesson without a **solution** folder & files the "Solve" button still appears and will set the file content to be empty.

### Steps to reproduce

1. Go to https://demo.tutorialkit.dev/forms-tutorial/introduction/welcome/
2. Click on the Solve button

### Expected behavior

There shouldn't be a solve button if the lesson does not have solution files. You should allow lessons that are exploratory without a specific solution. 

### Screenshots

![image](https://github.com/stackblitz/tutorialkit/assets/21179959/c3962952-8987-4946-81f7-b02f51d1b348)
^This lesson doesn't have a solution folder

![image](https://github.com/stackblitz/tutorialkit/assets/21179959/a1aab865-047a-4c94-92fa-43b6750783da)
^Solve button is still shows and clears the file content. 

### Platform

- TutorialKit version: [e.g. 1.0.1]
- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

Nemikolh:
Hey @leonyoung1! Thanks for the report!

Looks like a a bug! We should be setting the `helpAction` to `reset` in `WorkspacePanel.tsx`.

https://github.com/stackblitz/tutorialkit/blob/9805996a4211a1c8a3e1bfbbd958a27f1957d4d7/packages/components/react/src/Panels/WorkspacePanel.tsx#L89-L93

Would you be interested in contributing a fix?

leonyoung1:
Hi @Nemikolh, I have never contributed before but would love to.

Nemikolh:
@leonyoung1, amazing! :tada: 

Yes, create a pull request against this repo!

Nemikolh:
Closed by #108

PR #108, opened 2024-06-30
fix: problem with creating lessons without solution files

leonyoung1:
fix for https://github.com/stackblitz/tutorialkit/issues/107 In the hasSolution function of TutorialStore, `this._lessons.solution` is an array with the structure [lessonName, [list of solution files]]. Instead of checking the length of the solution array. we have to check the length of the second index which is the list of solution files.


Nemikolh (approved):
Good catch! :star_struck: 

And thanks a lot for the fix :tada:

PR #112, opened 2024-07-03
feat: support overriding `TopBar`

AriPerkkio:
Adds support for overriding `TopBar` component. In future we can add support for more component overrides.

```ts

export default defineConfig({
  integrations: [
    tutorialkit({
      components: {
        TopBar: './CustomTopBar.astro',
      },
    }),
  ],
});
```

Users may still want to utilize TutorialKit's default `<Logo>` and `<ThemeSwitch>`. These are accessible through slots:

```jsx
// ./CustomTopBar.astro
<nav style="display: flex; align-items: center; gap: 2rem; padding: 1rem;">
  Some content here
  <slot name="logo" />
  and here
  <slot name="theme-switch" />
  here too
  <slot name="login-button" />
  and here!
</nav>
```


docs/tutorialkit.dev/src/content/docs/guides/overriding-components.mdx
```diff
@@ -0,0 +1,62 @@
+---
+title: 'Overriding Components'
+description: "Override TutorialKit's default components to fit your needs."
+---
```
AriPerkkio, line 4: These docs are highly inspired by Starlight: https://starlight.astro.build/guides/overriding-components/

AriPerkkio:
I spent some time trying to set up tests for this kind of features but it seemed a bit bigger task than I expected. I don't think there are any existing tests where this feature could be easily tested..? Let's build good testing setup in separate PR later.

Next I'll look into adding overrides for other components so that rest of the feature requests can be filled. For example the bottom-left corners "Edit this page" link.

Nemikolh:
> Let's build good testing setup in separate PR later.

I totally agree. In particular I think this would be easy to test if we had playwright or puppeteer configured.

docs/tutorialkit.dev/src/content/docs/guides/overriding-components.mdx
```diff
@@ -0,0 +1,66 @@
...
+import { Image } from 'astro:assets';
+import uiTopBar from './images/ui-top-bar.png';
+
+TutorialKit's default components are customizable with [theming](/reference/theming/) options.
+In cases where theming is not enough you can override built-in components with your own ones.
```
Nemikolh, line 9:
```suggestion
In cases where theming is not enough you can override some built-in components with your own ones.
```

docs/tutorialkit.dev/src/content/docs/guides/overriding-components.mdx
```diff
@@ -0,0 +1,66 @@
...
+    tutorialkit({
+      components: {
+        // Component overrides go here
+      },
+    }),
+  ],
+});
+```
+
+For example to override `TopBar`, provide a path to the new component:
```
Nemikolh, line 30:
```suggestion
For example to override `TopBar`, provide a path to the new Astro component:
```

(just to emphasize that you cannot use React to write this component)

docs/tutorialkit.dev/src/content/docs/guides/overriding-components.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+```ts
+tutorialkit({
+  components: {
+    TopBar: './src/components/CustomTopBar.astro',
+  },
+}),
+```
+
+See [User Interface Reference](/guides/ui/) for component reference.
```
Nemikolh, line 40: Is that the correct link? This page describes the user interface but does not provide a reference.

package.json
```diff
@@ -13,7 +13,7 @@
     "demo": "pnpm run --filter=demo.tutorialkit.dev dev",
     "demo:build": "pnpm run --filter=demo.tutorialkit.dev build",
     "lint": "eslint \"{packages,docs}/**/*\"",
-    "test": "pnpm run --filter=@tutorialkit/* --filter=tutorialkit test"
+    "test": "pnpm run --stream --filter=@tutorialkit/* --filter=tutorialkit test --run"
```
Nemikolh, line 16: Oh interesting! :eyes: What do these flags do?

docs/tutorialkit.dev/src/content/docs/guides/overriding-components.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+### Top Bar
+
+<Image src={uiTopBar} alt="TutorialKit's Top bar" />
+
+When overriding `TopBar` you can place TutorialKit's default components using following [Astro slots](https://docs.astro.build/en/basics/astro-components/#named-slots):
+
+- `logo`: Logo of the application
+- `theme-switch`:  Switch for changing the theme
+- `login-button`: TODO, add description here
```
Nemikolh, line 50: This is only for enterprise users. So you could say something like:

```
For StackBlitz Enterprise user, the login button.
```

AriPerkkio:
> I totally agree. In particular I think this would be easy to test if we had playwright or puppeteer configured.

Yep! I had something like this in mind:

<details>

```ts
import { fileURLToPath } from "node:url";
import { beforeAll, expect, test } from "vitest";
import { type Browser, chromium } from "playwright";
import { dev } from "astro";

let browser: Browser;

beforeAll(async () => {
  // Start server
  const server = await dev({
    configFile: false,
    root: fileURLToPath(new URL("./fixtures", import.meta.url)),
    integrations: [
      tutorialkit({
        components: {
          TopBar: "./src/components/CustomTopBar.astro",
        },
      }),
    ],
  });

  // Open browser
  browser = await chromium.launch({ headless: true });

  // Close server
  return async function cleanup() {
    await server.stop();
  };
});

test("custom top bar is visible", async () => {
  const page = await browser.newPage();
  await page.goto("http://localhost:4321/");

  await page.getByRole("heading", {
    level: 1,
    name: "Custom top bar",
  });
});
```

</details>

But TutorialKit had some issues when used from Astro's Node API and changing `root`.

packages/astro/src/vite-plugins/override-components.ts
Nemikolh, line 1: Nice :pinched_fingers: Really nice comments!! :heart_eyes:

packages/astro/types.d.ts
```diff
@@ -7,3 +7,9 @@ declare module 'tutorialkit:store' {
 declare module 'tutorialkit:core' {
   export const webcontainer: import('@webcontainer/api').WebContainer;
 }
+
+declare module 'tutorialkit:override-components' {
+  const topBar: typeof import('./src/default/components/TopBar.astro').default;
+
+  export { topBar as TopBar };
+}
```
Nemikolh, line 15: This should be moved in `env-default.d.ts`. I don't think we mean to make that visible.

packages/astro/tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "moduleResolution": "Bundler",
     "outDir": "dist"
   },
-  "include": ["src"],
+  "include": ["src", "types.d.ts"],
```
Nemikolh, line 10: This can be removed once the declare module is moved to `env-default.d.ts`.

packages/astro/src/vite-plugins/override-components.ts
```diff
@@ -0,0 +1,72 @@
...
+   * ```
+   */
+  TopBar?: string;
+}
+
+const virtualModuleId = 'tutorialkit:override-components';
+const resolvedId = `\0${virtualModuleId}`;
+
+export function overrideComponents(options?: OverrideComponentsOptions): VitePlugin {
+  const topBar = options?.TopBar || '@tutorialkit/astro/default/components/TopBar.astro';
```
Nemikolh, line 51: This should also check whether `defaultRoutes` is false or not. If it's `false`, then we don't want to use the one from `@tutorialkit/astro` but the one from `src/components/TopBar.astro`.

This is needed to make sure that `eject` works as expected.

Note that I would agree that they should probably remove the import to `tutorialkit:override-components`, after doing `eject`. If you prefer you could also modify `eject` to do so.

Nemikolh (commented):
Looks pretty good and the docs are top notched!! :star_struck: 

Left a few comments, the more important being to consider the impact of doing an `eject`.

docs/tutorialkit.dev/src/content/docs/guides/overriding-components.mdx
AriPerkkio: That page describes what `Top Bar` is. But I've also added the same picture on this page so I guess we can remove this link completely.

Nemikolh: Yeah let's remove it

package.json
AriPerkkio: - `--run` is for Vitest to tell it to not run in watch-mode. Otherwise Vitest will stop after first package finishes run and waits for file changes. Other packages are not run at all until test run is cancelled.
- `--stream` is for pnpm to not collapse the output of the commands. 
  - <details>
    <summary>Before and after pictures</summary>
    
    No `--stream`:
    
    With `--stream`:
    

   </details>

packages/astro/src/vite-plugins/override-components.ts
AriPerkkio: I think it makes more sense to modify the import path in source files, rather than the plugin itself. I'll check how to do this during `eject` command.

AriPerkkio: I've added support for `tutorialkit eject` command now. During that we'll:
- Remove `components: {...}` from `astro.config.ts`
- Replace `tutorialkit:override-components` imports with relative file imports. If user had defined custom `TopBar` in `astro.config.ts` we'll use that one. If not, use default one. 

I'll be updating and adding test cases next.

AriPerkkio: > Can you elaborate why? The plugin is internal and the change is really simple, no?

To me it makes sense that imports from virtual module `tutorialkit:override-components` are removed during `tutorialkit eject`. If they are left as is, we would consider the internal Vite plugin as part of the public API. What do you think?

> I prefer the copy

We do still copy all the files as is. Now we just replace the import from virtual module to the actual file.

Nemikolh: > To me it makes sense that imports from virtual module `tutorialkit:override-components` are removed during `tutorialkit eject`. If they are left as is, we would consider the internal Vite plugin as part of the public API. What do you think?

Ah good point, I didn't thought about that.

It might not be too bad though given how simple the plugin is and I don't see it changing much. That plus the fact that the Astro integration cannot be removed after doing an `eject` as all the other plugins it adds are needed.

However with having the renaming on disk, now the name of the file becomes part of the public API which in my opinion is much harder to maintain.

> Now we just replace the import from virtual module to the actual file.

I think this is where it sounds great on paper, but long term it won't work well because:

 * Right now the CLI looks at the `@tutorialkit/astro/dist/default` folder from their project and just copy the files. We do this because we don't want to inject incompatible code and also because we don't _care_ about it's setup, we now it works when `defaultRoutes: true`, it will work when `defaultRoutes: false`.
 * The Astro integration might be a different version of the CLI, so maybe the CLI will forget to rename some components because it's too old, maybe it will crash thinking that a rename would work when it doesn't because the project is too old.
 * What if in the future we rename those files? Then the CLI will need to maintain all names in mind, and again if they use a CLI that's too old (which if they run `npx tutorialkit eject` would likely be the case) then it wouldn't work.

The simple copy is forward / backward compatible because it doesn't care about the internals. It just move files around in their project and makes no assumption about them.

AriPerkkio: > What if in the future we rename those files? Then the CLI will need to maintain all names in mind, and again if they use a CLI that's too old

We could resolve the name from `@tutorialkit/astro/default/components/TopBar.astro` entrypoint. Then it would remain up-to-date always. Removing that requires major version release as it's part of package's `exports`.

If we don't remove `tutorialkit:override-components` from source code that's generated to file system during `tutorialkit eject`, what's the alternative? Add the same resolving logic to the plugin? Something like below?

```diff
export function overrideComponents(options?: Options): VitePlugin {
+  // Fallback to local relative file when `defaultRoutes` is disabled, otherwise use public entrypoint
+  const fallback = options?.defaultRoutes === false ? './TopBar.astro' : '@tutorialkit/astro/default/components/TopBar.astro';
+
+  const topBar = options?.components?.TopBar || fallback;
-  const topBar = options?.components?.TopBar || '@tutorialkit/astro/default/components/TopBar.astro';
```

Nemikolh: > We could resolve the name from `@tutorialkit/astro/default/components/TopBar.astro` entrypoint. Then it would remain up-to-date always. Removing that requires major version release as it's part of package's `exports`.

That would help a bit but it wouldn't solve the CLI / project version mistmatch. You would still need to use the same version between the two or it would break.

I suppose we could also throw an error if the version of `tutorialkit` in the project does not match the version of the CLI or install the correct CLI and `execve` into that.

> If we don't remove `tutorialkit:override-components` from source code that's generated to file system during `tutorialkit eject`, what's the alternative? Add the same resolving logic to the plugin? Something like below?

Yes, this is what I had in mind (or if it doesn't work use the `config.projectRoot` to resolve the path to an absolute path)

packages/astro/src/vite-plugins/override-components.ts
```diff
@@ -0,0 +1,92 @@
+/**
+ * A plugin that lets users to override TutorialKit's components.
```
d3lm, line 2:
```suggestion
 * A plugin that lets users override TutorialKit's components.
```

packages/astro/src/vite-plugins/override-components.ts
```diff
@@ -0,0 +1,92 @@
...
+import type { VitePlugin } from '../types.js';
+
+export interface OverrideComponentsOptions {
+  /**
+   * Component for overriding the top bar.
+   *
+   * This component has 3 slots that are used to pass TutorialKit's default components:
+   * - `logo`: Logo of the application
+   * - `theme-switch`:  Switch for changing the theme
+   * - `login-button`: For StackBlitz Enterprise user, the login button
```
d3lm, line 36: I'd indent the bullet points with 2 spaces.

packages/astro/src/vite-plugins/override-components.ts
```diff
@@ -0,0 +1,92 @@
...
+   *
+   * Usage:
+   * ```jsx
```
d3lm, line 39: Nit and personal preference but I'd add a newline before.

AriPerkkio:
Tests on Windows CI were failing but it's fixed now. ✅

Nemikolh (approved):
Nice, really good work on this!! :star_struck:

PR #119, opened 2024-07-09
ci(publish): fix publish config

AriPerkkio:
- Scoped packages require `--access public`. It's designed to prevent accidentally publishing scoped packages to public. We do it intentionally for `@tutorialkit/*` packages.
- Github Actions instruct to set `registry` explicitly, though I'm not sure if this is required. 
  - > Please note that you need to set the `registry-url` to `https://registry.npmjs.org/` in `setup-node` to properly configure your credentials.
    > https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages
- `--provenance` flag requires `repository` in `package.json`

Example run which published couple of packages:
- https://github.com/stackblitz/tutorialkit/actions/runs/9854149942/job/27206280839
- https://www.npmjs.com/package/@tutorialkit/runtime

⚠️ Once this is merged, we should immediately publish next `0.0.1-alpha.25`. It's fine to skip `24` for the other packages.

### DIFF
--- a/.github/actions/setup-and-build/action.yml
+++ b/.github/actions/setup-and-build/action.yml
@@ -14,6 +14,7 @@ runs:
       uses: actions/setup-node@v4
       with:
         node-version: ${{ inputs.node-version }}
+        registry-url: 'https://registry.npmjs.org'
 
     - name: Install pnpm
       uses: pnpm/action-setup@v4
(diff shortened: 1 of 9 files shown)


Nemikolh (approved):
Amazing find!! :star2:

PR #124, opened 2024-07-09
fix(ci): use PAT to make sure release PRs trigger workflows

Nemikolh:
This PR replace the default use of the `GITHUB_TOKEN` with `GITOPS_REPO_PAT` which means that release PRs should now properly trigger workflows.

### DIFF
--- a/.github/workflows/prepare-release.yaml
+++ b/.github/workflows/prepare-release.yaml
@@ -31,7 +31,7 @@ jobs:
         run: pnpm run changelog
 
       - name: Create Pull Request
-        uses: peter-evans/create-pull-request@v6
+        uses: peter-evans/create-pull-request@c5a7806660adbe173f04e3e038b0ccdcd758773c
         with:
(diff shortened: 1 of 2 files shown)


AriPerkkio (changes requested):
We need same here:

https://github.com/stackblitz/tutorialkit/blob/3ccfcc2ba050259c28ae945ea5b85f748b5956d5/.github/workflows/prepare-release.yaml#L33-L41

AriPerkkio (approved):
Looks good to me!

One thing that comes to mind though - should we lock version of the Github action to specific commit? I'm always a bit worried when passing secrets to custom Github Actions that use tag as version. So something like:

```yaml
 - name: Create Pull Request
   uses: peter-evans/create-pull-request@<git-commit-hash-here>
```

AriPerkkio (approved):

AriPerkkio:
The token is missing: https://github.com/stackblitz/tutorialkit/actions/runs/9908586698/job/27374683334

@Nemikolh @apai4

PR #130, opened 2024-07-11
feat: add "Edit this page" link

AriPerkkio:
Adds support for "Edit this page" link.

```
---
type: part
title: Basics
editPageLink: https://github.com/stackblitz/tutorialkit/tree/main/packages/template/src/content/tutorial/:path
---
```

Inherited `editPageLink` can be disabled on lower hierarcy:

```
---
type: lesson
title: Some lesson
editPageLink: false
---
```

See bottom-left corner:


docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx
```diff
@@ -170,3 +177,30 @@ Navigating to a lesson that specifies `autoReload` will always reload the previe
 Specified which folder from the `src/templates/` directory should be used as the basis for the code. See the "[Code templates](/guides/creating-content/#code-templates)" guide for a detailed explainer.
 <PropertyTable inherited type="string" />
 
+#### `editPageLink`
+Display a link in lesson for editing the page content.
+The value is a URL pattern where `:path` is replaced with the lesson's location relative to the `src/content/tutorial`.
```
AriPerkkio, line 182: @Nemikolh is this location logic OK? It's using the `id` from Astros `entry.id` now.

Nemikolh: I think this is fine yeah. :+1:

docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx
```diff
@@ -48,6 +48,13 @@ type I18nText = {
    */
   partTemplate?: string,
 
+  /**
+   * Text of the edit page link.
+   *
+   * @default 'Edit this page'
+   */
+  editPage?: string
```
Nemikolh, line 56: Nit: `editPage` -> `editPageText`.

The idea is:
 - When it ends with `Text` there are no interpolation, the text is copied verbatim.
 - When it ends with `Template`, string interpolation is possible.

packages/types/src/schemas/common.ts
```diff
@@ -153,9 +153,18 @@ export const webcontainerSchema = commandsSchema.extend({
...
+  z.string(),
+
+  // `false` for disabling the edit link
+  z.boolean(),
+]);
+
 export const baseSchema = webcontainerSchema.extend({
   title: z.string(),
   slug: z.string().optional(),
+  editPageLink: editPageLinkSchema.optional(),
```
Nemikolh, line 167: This should be set inside the (incorrectly named now) `webcontainerSchema` otherwise you can't set the property at the tutorial level which is a bit annoying.

packages/types/src/schemas/i18n.ts
```diff
@@ -8,6 +8,13 @@ export const i18nSchema = z.object({
    */
   partTemplate: z.string().optional(),
 
+  /**
+   * Text of the edit page link.
+   *
+   * @default 'Edit this page'
+   */
+  editPage: z.string().optional(),
```
Nemikolh, line 16: `editPage` -> `editPageText`

packages/types/src/schemas/tutorial.ts
```diff
@@ -1,9 +1,10 @@
 import { z } from 'zod';
-import { webcontainerSchema } from './common.js';
+import { webcontainerSchema, editPageLinkSchema } from './common.js';
 
 export const tutorialSchema = webcontainerSchema.extend({
   type: z.literal('tutorial'),
   logoLink: z.string().optional(),
+  editPageLink: editPageLinkSchema.optional(),
```
Nemikolh, line 7: This can be removed once the property is in the `webcontainerSchema`.

packages/astro/src/default/utils/content.ts
```diff
@@ -274,9 +287,11 @@ export async function getTutorial(): Promise<Tutorial> {
         href: joinPaths(baseURL, `/${partSlug}/${chapterSlug}/${nextLesson.slug}`),
       };
     }
-  }
 
-  // console.log(inspect(_tutorial, undefined, Infinity, true));
+    if (lesson.data.editPageLink && typeof lesson.data.editPageLink === 'string') {
+      lesson.editPageLink = lesson.data.editPageLink.replace(':path', lesson.filepath);
```
Nemikolh, line 292: You might need to export the `interpolateString` utility from `@tutorialkit/components-react` and then use it here.

Actuaclly, I'm starting to think that we should move the `interpolateString` utility in the `@tutorialkit/types` package. Wdyt?

Nemikolh (commented):
This is looking great! :smiley: 

One immediate feedback that comes to mind is regarding the use of `:path`. I think I see where this is coming from but I would prefer if we used `${path}` because:

 * It's not behaving as route parameters in express-like servers where `:path` would only match a segment (so only until the next `/`). Here it instead does a string replace.
 * To use the same convention everywhere. If we're sold on using `${variable}` for string interpolation in the metadata then let's use it everywhere we need that feature.

Note that I'm also open to a different syntax. I picked `${variable}` initially because it's similar to JavaScript's template literal syntax.

AriPerkkio:
> One immediate feedback that comes to mind is regarding the use of `:path`. I think I see where this is coming from but I would prefer if we used `${path}` because:

I've changed the pattern to `${path}` now - it's better to be consistent with rest of the codebase. And `partTemplate` was already following this pattern. The `:path` convention was from VitePress.

Nemikolh (approved):
Nice!! Really good work on this :star_struck: 

Love it :heart:

PR #127, opened 2024-07-10
fix: mobile fixes and basic i18n support

Nemikolh:
This PR:
 * Makes a few changes to the rendering of the navigation dropdown on mobile (arrows on both sides are removed).

comparison-before-after

 * Fixes a couple of issues on the mobile version (like buttons not visible).
 * Adds transitions between dark / light themes.
 * Shows a page loading indicator when navigation between two page is greater than 500ms.

[loading-indicator.webm](https://github.com/stackblitz/tutorialkit/assets/4008156/a2b80e01-5c00-4b88-b87c-16882946e35c)

 * Adds a button to start the tutorial on platform where memory is more restricted.
 * Introduces `i18n` in the metadata that you typically want to configure on the tutorial level to change the text of the UI:

```md
---
type: tutorial
i18n:
  partTemplate: Partie ${index}]: ${title}
  noPreviewNorStepsText: Pas d'aperçu ou d'étapes à montrer
  startWebContainerText: Lancer le tutoriel
---
```

 * Implements deep inheritance for the metadata for properties like `terminal` and `i18n` so that not every property must be specified again if only one changed. Note that this only works for objects. For arrays, string and numbers the value is always overriden (which is typically what you'd want from TutorialKit's metadata).

Issue #128, opened 2024-07-10
Watch mode for development


AriPerkkio:
### Is your feature request related to a problem?

Developing TutorialKit requires running `pnpm build` constantly after changes. Having automatic builds on file change would reduce required manual steps. 

### Describe the solution you'd like.

Add `dev` command into root of monorepo that would run `dev` command in every single package. Each package should have their own `dev` command that runs their specific build tool (`tsc`, `esbuild`) in watch mode. 

Ideally `packages/template` would also restart the Astro server on monorepo dependency changes.

PR #133, opened 2024-07-12
feat: finalize basic i18n support

Nemikolh:
This PR completes the initial i18n support added in #127 for the remaining part of the UI where text should be customizable.


AriPerkkio (approved):
Looks good to me!

In future we'll need to be careful with the `packages/components/react` package. Exported public components should not define any texts that need localizing. They should always accept them as props.

Nemikolh:
> Exported public components should not define any texts that need localizing. They should always accept them as props.

That's a very good point. Maybe we should write that down in a `README.md` or `CONTRIBUTING.md`.

What do you think?

AriPerkkio:
That would be great. I've been planning to create good `CONTRIBUTING.md` once #128 is done. 

For ESLint I've used this kind of rule before: https://gist.github.com/AriPerkkio/fab0f24753b0f8457b63da187a4aee9e. That restricts importing from specific directories from files that inside specific directories. E.g. prevent importing from `packages/admin` in files that are inside `packages/user`.

Nemikolh:
Oh interesting! I like a lot the idea of using eslint to enforce this.

From a quick look around, this eslint rule came up: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-literals.md

It's not perfect but it would caught a few usecase.

AriPerkkio (approved):

PR #139, opened 2024-07-15
fix: a transition-theme class was missing for the content right border

Nemikolh:
This PR adds a missing `transition-theme` class on the border. Thanks @d3lm for the find! :smiley:

### DIFF
--- a/packages/astro/src/default/components/MainContainer.astro
+++ b/packages/astro/src/default/components/MainContainer.astro
@@ -34,7 +34,10 @@ const showWorkspacePanel = hasWorkspace(lesson);
     <Nav client:load lesson={lesson} navList={navList} />
     <TutorialContent lesson={lesson} />
   </div>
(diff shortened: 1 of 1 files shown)


AriPerkkio (approved):
Looks good. I wonder if there's an easier way to add this class everywhere. Maybe its rules should automatically be applied on each color class. 🤔

d3lm (approved):

Nemikolh:
@AriPerkkio I think we could definitely do that and I'm torn a bit on that one.

I was thinking of doing that at first but I changed my mind mainly because it's less explicit and it can have unintended side effects. For instance on components added by end users where maybe they want to control the transition and can't easily because of our plugin.

It also might have a perf impact if it's on everything but it would have some annoying side effects.

Instead of answering all of those questions, I took the simplest and more explicit approach. I don't think it's too bad tbqh.

PR #140, opened 2024-07-15
test: enable "build project" test cases

AriPerkkio:
- Enable the test cases that required `@tutorialkit/theme` to be published

https://github.com/AriPerkkio/tutorialkit/actions/runs/9974724567/job/27562811019


packages/template/package.json
```diff
@@ -25,7 +25,7 @@
     "@types/node": "^20.14.6",
     "@types/react": "^18.3.3",
     "@unocss/reset": "^0.59.4",
-    "astro": "4.10.3",
+    "astro": "^4.10.3",
```
AriPerkkio, line 28: This is required for `yarn`. Without this it would install two copied of `astro` and complain about type conflict in `astro.config.ts`.

Nemikolh: Oh good catch!

Nemikolh (approved):
Good catch for this test! :smiley:  I had forgotten about it :sweat_smile:

PR #141, opened 2024-07-15
chore: watch mode

AriPerkkio:
- Fixes https://github.com/stackblitz/tutorialkit/issues/128

### DIFF
--- a/.github/commit-convention.md
+++ b/.github/commit-convention.md
@@ -1,0 +1,92 @@
+## Git Commit Message Convention
+
+> This is adapted from [Angular's commit convention](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-angular).
+
+#### TL;DR:
+
+Messages must be matched by the following regex:
+
+<!-- prettier-ignore -->
+```js
+/^(revert: )?(feat|fix|docs|dx|refactor|perf|test|workflow|build|ci|chore|types|wip|release|deps)(\(.+\))?: .{1,50}/
+```
+
+#### Examples
+
+Appears under "Features" header, `dev` subheader:
+
+```
+feat(dev): add 'comments' option
+```
+
+Appears under "Bug Fixes" header, `dev` subheader, with a link to issue #28:
+
+```
+fix(dev): fix dev error
+
+close #28
+```
+
+Appears under "Performance Improvements" header, and under "Breaking Changes" with the breaking change explanation:
+
+```
+perf(build): remove 'foo' option
+
+BREAKING CHANGE: The 'foo' option has been removed.
+```
+
+The following commit and commit `667ecc1` do not appear in the changelog if they are under the same release. If not, the revert commit appears under the "Reverts" header.
+
+```
+revert: feat(compiler): add 'comments' option
+
+This reverts commit 667ecc1654a317a13331b17617d973392f415f02.
+```
+
+### Full Message Format
+
+A commit message consists of a **header**, **body** and **footer**. The header has a **type**, **scope** and **subject**:
+
+```
+<type>(<scope>): <subject>
+<BLANK LINE>
+<body>
+<BLANK LINE>
+<footer>
+```
+
+The **header** is mandatory and the **scope** of the header is optional.
+
+### Revert
+
+If the commit reverts a previous commit, it should begin with `revert: `, followed by the header of the reverted commit. In the body, it should say: `This reverts commit <hash>.`, where the hash is the SHA of the commit being reverted.
+
+### Type
+
+If the prefix is `feat`, `fix` or `perf`, it will appear in the changelog. However, if there is any [BREAKING CHANGE](#footer), the commit will always appear in the changelog.
+
+Other prefixes are up to your discretion. Suggested prefixes are `docs`, `chore`, `style`, `refactor`, and `test` for non-changelog related tasks.
+
+### Scope
+
+The scope could be anything specifying the place of the commit change. For example `dev`, `build`, `workflow`, `cli` etc...
+
+### Subject
+
+The subject contains a succinct description of the change:
+
+- use the imperative, present tense: "change" not "changed" nor "changes"
+- don't capitalize the first letter
+- no dot (.) at the end
+
+### Body
+
+Just as in the **subject**, use the imperative, present tense: "change" not "changed" nor "changes".
+The body should include the motivation for the change and contrast this with previous behavior.
+
+### Footer
+
+The footer should contain any information about **Breaking Changes** and is also the place to
+reference GitHub issues that this commit **Closes**.
+
+**Breaking Changes** should start with the word `BREAKING CHANGE:` with a space or two newlines. The rest of the commit message is then used for this.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,10 +50,10 @@ And re-run `pnpm install` to link the package.
 
 - If adding a new feature:
 
-  - Add accompanying test case.
   - Provide a convincing reason to add this feature. Ideally, you should open a suggestion issue first and have it approved before working on it.
+  - Add accompanying test case.
 
-- If fixing bug:
+- If fixing a bug:
 
   - If you are resolving a special issue, add `(fix #xxxx[,#xxxx])` (#xxxx is the issue id) in your PR title for a better release log, e.g. `fix: update entities encoding/decoding (fix #3899)`.
   - Provide a detailed description of the bug in the PR. Live demo preferred.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,66 @@
+# TutorialKit Contributing Guide
+
+Hi! We are really excited that you are interested in contributing to TutorialKit. Before submitting your contribution, please make sure to take a moment and read through the following guide:
+
+## Repo Setup
+
+The TutorialKit repo is a monorepo using pnpm workspaces. The package manager used to install and link dependencies must be [pnpm](https://pnpm.io/).
+
+To develop and test packages:
+
+1. Clone this repository and navigate into the cloned directory.
+
+```
+git clone git@github.com:stackblitz/tutorialkit.git
+cd tutorialkit
+```
+
+2. Run `pnpm install` in project's root folder
+
+3. Run `pnpm run build` to build sources
+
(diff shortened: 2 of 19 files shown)


AriPerkkio:
There's still an issue on `@tutorialkit/astro > @tutorialkit/react-components` HMR. The `packages/template` sees the update of `@tutorialkit/react-components` and HMR kicks in, but it's served from cached optimized deps from `.vite/deps`. Restarting Vite/Astro server of `packages/template` with `--force` works.

I've tried all kinds of combinations of `optimizeDeps.exclude` but so far nothing works.

CONTRIBUTING.md
```diff
@@ -0,0 +1,58 @@
...
+  "pnpm": {
+    "overrides": {
+      "@tutorialkit/astro": "file:../tutorialkit/packages/astro",
+      "@tutorialkit/components-react": "file:../tutorialkit/packages/components/react",
+      "@tutorialkit/runtime": "file:../tutorialkit/packages/runtime",
+      "@tutorialkit/theme": "file:../tutorialkit/packages/theme",
+      "@tutorialkit/types": "file:../tutorialkit/packages/types"
+    }
+  }
+}
```
AriPerkkio, line 34: This works somewhat ok but is quite far from ideal. Due to `file:` usage, the project must re-run `pnpm install` after TutorialKit packages have updated.

Using `link:` make changes automatically visible, but I was unable to solve issue where TutorialKit's dependencies brought multiple copies of React into the run time and caused hooks to fail. With webpack this could be solved using `resolve.alias` but it just doesn't work with Vite.

CONTRIBUTING.md
```diff
@@ -0,0 +1,58 @@
...
+
+The TutorialKit repo is a monorepo using pnpm workspaces. The package manager used to install and link dependencies must be [pnpm](https://pnpm.io/).
+
+To develop and test packages:
+
+1. Run `pnpm install` in project's root folder
+
+2. Run `pnpm run build` to build sources
+
+3. Run `pnpm run dev` to build sources in watch mode
```
AriPerkkio, line 15: This works fine for most of the packages. Changes of `packages/components/react` are showing up on `packages/template` randomly.

Though sometimes `tsc` builds stop completely working and `scripts/clean.sh` is required. Personally I would rather remove TS project references as they cause hard to debug issues in this project.

AriPerkkio:
> There's still an issue on `@tutorialkit/astro > @tutorialkit/react-components` HMR. 

Adding `optimizeDeps` on `packages/astro` seems to fix it - exactly as `TUTORIALKIT_DEV` does! 🥳

I tried to force it from the `packages/template` using `@tutorialkit/astro > @tutorialkit/react-components` in `optimizeDeps`. It should work, in theory at least.

packages/astro/scripts/build.js
```diff
@@ -1,40 +1,85 @@
...
-if (existsSync('./dist/default')) {
-  assert.fail('TypeScript transpiled the default folder, it means that the tsconfig has an issue');
+const isWatch = process.argv.includes('--watch');
+
+await generateTypes();
+await buildJS();
+await copyDefaultFolder();
+
+async function generateTypes() {
+  if (!isWatch) {
```
Nemikolh, line 17: Nit: I would always delete `dist`. It's too easy after a rename to have a file staying around and causing issues.

packages/astro/scripts/build.js
```diff
@@ -1,40 +1,85 @@
...
+    isWatch && '--watch',
+    '--preserveWatchOutput',
+  ].filter((s) => !!s);
+
+  const promise = execa('tsc', args, { stdio: 'inherit', preferLocal: true });
+
+  if (!isWatch && existsSync('./dist/default')) {
+    await promise;
+    assert.fail('TypeScript transpiled the default folder, it means that the tsconfig has an issue');
+  }
```
Nemikolh, line 36: That's pretty nice :pinched_fingers:

packages/astro/scripts/build.js
```diff
@@ -1,40 +1,85 @@
...
+const isWatch = process.argv.includes('--watch');
+
+await generateTypes();
+await buildJS();
+await copyDefaultFolder();
+
+async function generateTypes() {
+  if (!isWatch) {
+    // clean dist
+    await rm('dist', { recursive: true, force: true });
```
Nemikolh, line 19: Maybe this should be done outside of this function. It doesn't feels specific to generating the types.

packages/astro/scripts/build.js
```diff
@@ -1,40 +1,85 @@
...
+
+async function copyDefaultFolder() {
+  const src = './src/default';
+  const dist = './dist/default';
+  await rm(dist, { recursive: true, force: true });
+
+  // copy default folder unmodified, without test files
+  await cp(src, dist, {
+    recursive: true,
+    filter: (filename) => !filename.endsWith('.spec.ts'),
```
Nemikolh, line 69: Oh this is really neat :star2:

packages/cli/package.json
```diff
@@ -22,6 +22,7 @@
   "scripts": {
     "build": "node scripts/build.js",
     "build-release": "node scripts/build-release.js",
+    "dev": "node scripts/build.js --watch",
```
Nemikolh, line 25: I don't think having a watch mode for the CLI makes sense. You have to run the CLI via the command line, and it's pretty simple to add a `pnpm run build` in front of it before running it.

The cost on memory usage is not worth it IMO.

packages/components/react/scripts/build.js
Nemikolh, line 1: This is super clean, really like those changes :tada:

packages/create-tutorial/package.json
```diff
@@ -17,6 +17,7 @@
   },
   "scripts": {
     "build": "rm -rf dist && tsc -b",
+    "dev": "tsc -b --watch --preserveWatchOutput",
```
Nemikolh, line 20: Same reason as for the CLI, I would remove this script.

packages/test-utils/package.json
```diff
@@ -11,7 +11,8 @@
     "dist"
   ],
   "scripts": {
-    "build": "tsc -b"
+    "build": "tsc -b",
+    "dev": "pnpm run build --watch --preserveWatchOutput"
```
Nemikolh, line 15: I'm not a fan of this either as it feels like 90% of the time you don't need it.

What if instead we had a `test:dev` for the `runtime` which ran `pnpm run --filter=test-utils build --watch --preserveWatchOutput`?

packages/types/package.json
```diff
@@ -21,6 +21,7 @@
   ],
   "scripts": {
     "build": "tsc -b tsconfig.build.json",
+    "dev": "pnpm run build --preserveWatchOutput",
```
Nemikolh, line 24:
```suggestion
    "dev": "pnpm run build --watch --preserveWatchOutput",
```

Nemikolh (commented):
Looking good! I really like the improvement and normalization on the build script.

I would reduce a bit on the number of watch processes started though, I left some comments here and there to clarify where I think they could be removed.

.github/commit-convention.md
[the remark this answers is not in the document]
AriPerkkio: (FYI this is from Vitest repo)

README.md
```diff
@@ -31,29 +31,12 @@ Read our documentation on [tutorialkit.dev](https://tutorialkit.dev/guides/about
 
 ### Prerequisites
 
-- Install [Node.js](https://nodejs.org/en) v18.18 or above.
```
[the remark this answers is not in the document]
AriPerkkio: Oops, nope! I'll revert it. 🙌

packages/astro/scripts/build.js
AriPerkkio: In `pnpm` workspace this is required. When `--recursive --parallel` starts the scripts, we must have initial builds on file system ready. Otherwise the packages will try to build at the same time, and as their dependencies (other monorepo packages) are not present, they fail the build. https://egoist.dev/pnpm-monorepo

packages/cli/package.json
AriPerkkio: True - I'll remove this.

packages/test-utils/package.json
AriPerkkio: Actually this package should be consumed as source files by other packages. This package shouldn't have build step at all.

packages/astro/scripts/build.js
Nemikolh: Ah maybe we need something similar to what we've done in WebContainer then. In there, we wait for the `dist` folder to be ready before starting the `watch` tasks of the downstream packages.

For now, let's keep your logic then :+1:

packages/test-utils/package.json
Nemikolh: Oh I agree! That would be even better! :100: 

Do you think you could implement this in this PR?

CONTRIBUTING.md
AriPerkkio: > Without project references the DX of navigating the code in this mono-repo becomes much worse.

You mean navigating in IDE by clicking the lines? That doesn't required project references, just `paths`.

packages/test-utils/package.json
AriPerkkio: Applied in https://github.com/stackblitz/tutorialkit/pull/141/commits/483e068cdf4202e47b62de2b3b9fbda3fd376fe4 - what do you think?

Nemikolh: Oh this looks really great! :star_struck: Good job on this!! :muscle:

CONTRIBUTING.md
Nemikolh: Don't `paths` need to duplicate the export logic from the package?

I might be remembering it wrong.

Maybe there's a better way to configure all this than by using project references. I would be tempted to keep it though because this is what we use in other repositories.

AriPerkkio: Yep, it does need: https://github.com/vitest-dev/vitest/blob/32c0d99b3cd03df5570c6886b42498aa6a378fa2/tsconfig.base.json#L7-L34

Nemikolh: Ah ok, thanks for checking. Maybe we could move to that then, it doesn't seem so bad.

I think we should add that as an item to discuss to make sure everyone is on board with the idea.

Nemikolh (approved):
Let's goooo!!!

PR #145, opened 2024-07-16
fix(ci): override @astrojs/language-server with an older version

Nemikolh:
This PR fixes the error encountered in CI after the latest update to `@astrojs/language-server`. When we don't use file dependencies there are no issue. Running `astro check` on the template works completely fine and on a new project that use our npm deps it also works.

My guess is that our use of `file://` dependency in our test is kinda broken because some dependencies that should be shared aren't. I haven't been able exactly to put my finger on which one but having duplicate of the same dependency even if they are at the same version (here `@astrojs/language-server`) didn't work.

Note that this PR also updates a few dependencies and dedupe some of them (like the `prettier-plugin-astro` one).


packages/cli/src/commands/eject/index.ts
```diff
@@ -18,7 +18,7 @@ interface PackageJson {
 }
 
 const TUTORIALKIT_VERSION = pkg.version;
-const REQUIRED_DEPENDENCIES = ['@tutorialkit/runtime', '@webcontainer/api', 'nanostores', '@nanostores/react'];
+const REQUIRED_DEPENDENCIES = ['@tutorialkit/runtime', '@webcontainer/api', 'nanostores', '@nanostores/react', 'kleur'];
```
Nemikolh, line 21: After moving the CLI test from `npm` to `pnpm` it revealed a bug in our `eject` command. :tada:

packages/cli/tests/create-tutorial.test.ts
```diff
@@ -216,20 +193,39 @@ test('cannot eject on an astro project that is not using TutorialKit 2', async (
...
   return filePath.replace(/\\/g, '/');
 }
+
+async function runPnpmInstall(dest: string, baseDir: string) {
+  const packageJsonPath = path.join(dest, 'package.json');
+  const packageJson = JSON.parse(await fs.readFile(packageJsonPath, 'utf-8'));
+
+  packageJson.pnpm = {
+    overrides: {
+      '@astrojs/language-server': '2.11.1',
```
Nemikolh, line 217: This is the version that we were using before that didn't exhibit that issue. I still think that doing the `file:..` thing below is wrong. I'm wondering if moving to use `.tar.gz` files would solve the problem because then their `node_modules` would be entirely managed by the package manager.

pnpm-lock.yaml
```diff
@@ -601,52 +601,31 @@ packages:
...
-      fast-glob: 3.3.2
-      kleur: 4.1.5
-      typescript: 5.4.5
-      yargs: 17.7.2
-    transitivePeerDependencies:
-      - prettier
-      - prettier-plugin-astro
-    dev: true
-
-  /@astrojs/compiler@1.8.2:
```
Nemikolh, line 638: We now also have a single version of `@astrojs/compiler` instead of 2

packages/cli/tests/create-tutorial.test.ts
```diff
@@ -73,24 +73,25 @@ test('create and eject a project', async (context) => {
...
+  if (process.platform !== 'win32') {
+    await fs.rm(path.join(dest, 'node_modules'), { force: true, recursive: true, maxRetries: 5 });
+  }
 
-  const projectFiles = await fs.readdir(dest, { recursive: true });
+  let projectFiles = await fs.readdir(dest, { recursive: true });
+
+  if (process.platform === 'win32') {
+    projectFiles = projectFiles.filter((filePath) => !filePath.startsWith('node_modules'));
+  }
```
Nemikolh, line 94: I really don't like this hack but I couldn't find any other better way to remove this folder on Windows.

It seems that the issue is that when we try to delete the `node_modules`, we also delete `esbuild.exe` which get an `EBUSY` error. If we use the `maxRetry` property for that usecase, it times out.

My guess is that at this point there's still an `esbuild` process running. However, I don't know what spawned that `esbuild` process and I don't want to try to randomly kill all `esbuild` processes which would be way worse.

packages/cli/tests/create-tutorial.test.ts
```diff
@@ -23,9 +30,6 @@ test('cannot create project without installing but with starting', async (contex
   await expect(
     execa('node', [cli, 'create', name, '--no-install', '--start'], {
       cwd: tmpDir,
-      env: {
-        TK_DIRECTORY: baseDir,
-      },
```
AriPerkkio, line 28: Now that `TK_DIRECTORY` is not used at all, we could remove it from `packages/cli/src/utils/workspace-version.ts` completely.

Nemikolh: Oh yeah good point!

AriPerkkio (approved):
It's kinda hacky but package managers and node's resolving is confusing some times. I looked at the error locally quickly and didn't find better solution. I was comparing `pnpm-lock.yaml` of the generated project and locally linked packages and didn't see any issues. 🤷 

These changes also made the think that we should add post-release test cases that are pure e2e tests using packages from NPM.

PR #146, opened 2024-07-17
ci: separate `create-tutorial` releasing

AriPerkkio:
Separates `create-tutorial` package releasing from all other `@tutorialkit/*` and `tutorialkit` packages. The `create-tutorial` is essentially only an alias for installing `tutorialkit` package and calling it. I don't think we'll be publishing many versions of this package. 

- Added workflow for separately specifying version for `create-tutorial` and publishing it
- Defined `create-tutorial` package's dependency on `tutorialkit` as `latest`, so that package managers will always install the latest version of it.

Example:
- Manually run the preparation PR workflow: https://github.com/AriPerkkio/tutorialkit/actions/runs/9973462056
- PR is created: https://github.com/AriPerkkio/tutorialkit/pull/36
- Once PR is merged, publish starts automatically: https://github.com/AriPerkkio/tutorialkit/actions/runs/9973515249

### DIFF
--- a/.github/workflows/publish-create-tutorial.yaml
+++ b/.github/workflows/publish-create-tutorial.yaml
@@ -1,0 +1,66 @@
+name: Prepare create-tutorial Release PR
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version to publish, e.g. 0.0.1'
+        required: true
+        default: '0.0.1'
+        type: string
+  push:
+    branches:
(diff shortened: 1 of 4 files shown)


AriPerkkio:
I'll be publishing `0.0.1` of `@tutorialkit/*` and `tutorialkit` packages once this is merged. I think we could publish `3.0.0` (first available version) of `create-tutorial` then too.

PR #153, opened 2024-07-18
feat!: rename `tutorialkit` to `@tutorialkit/cli`

AriPerkkio:
- Renames npm package `tutorialkit` to `@tutorialkit/cli`
- Adds some documentation to `CONTRIBUTING.md` about package naming convention. This is mostly related to development time communication practices and doesn't affect end users.
- Renames (internally) `tutorialkit-vscode` to `tutorialkit`. Removes work-around from build script that was used to do this renaming right before publishing. This name is already used by the published extension. 

BREAKING CHANGES:
- `tutorialkit` npm package is now unused. We should probably deprecate that from npm? Or just leave it as is. 
- Adds new package `@tutorialkit/cli`. This package does not yet exist in npm. I'm not sure if publishing this package will work without any permission settings. Do we need to add https://www.npmjs.com/~stackblitz-devops there some how?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,7 +39,7 @@ These packages will be installed by the end-users in their `package.json`.
 ### CLI Packages
 
 The CLI packages are expected to be run by users using `npm create tutorial` and `npx tutorialkit` commands.
-These should not be installed in their `package.json`.
+These should not be added to `package.json`.
 
 - `@tutorialkit/cli`
 - `create-tutorial`
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,6 +24,26 @@ cd tutorialkit
 
 5. Run `pnpm run test` to run core tests
 
+The monorepo consists of multiple packages that are grouped into following groups:
+
+### Core Packages
+
+These packages will be installed by the end-users in their `package.json`.
+
+- `@tutorialkit/astro`
+- `@tutorialkit/components-react`
+- `@tutorialkit/runtime`
+- `@tutorialkit/theme`
+- `@tutorialkit/types`
+
+### CLI Packages
+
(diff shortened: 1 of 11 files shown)


Nemikolh:
> Adds new package `@tutorialkit/cli`. This package does not yet exist in npm

This package used to exist, I think we might have cleaned the alpha versions that we had published for testing. Hopefully that won't be a problem.

CONTRIBUTING.md
```diff
@@ -24,6 +24,26 @@ cd tutorialkit
...
+- `@tutorialkit/astro`
+- `@tutorialkit/components-react`
+- `@tutorialkit/runtime`
+- `@tutorialkit/theme`
+- `@tutorialkit/types`
+
+### CLI Packages
+
+The CLI packages are expected to be run by users using `npm create tutorial` and `npx tutorialkit` commands.
+These should not be installed in their `package.json`.
```
Nemikolh, line 42: Nit:
```suggestion
These should not be added to `package.json`.
```

README.md
```diff
@@ -36,7 +36,7 @@ Read our documentation on [tutorialkit.dev](https://tutorialkit.dev/guides/about
 
 ### Contribution
 
-See [Contributing Guide](https://github.com/stackblitz/tutorialkit/blob/main/CONTRIBUTING.md).
+See [Contributing Guide](./CONTRIBUTING.md).
```
Nemikolh, line 39: Nice 🤩

Nemikolh (approved):
Looks good! Love the cleanup as well 🤗

AriPerkkio:
I think this is actually not a breaking change as CLI packages are not installed by end-users. I'll resolve the conflicts and then we can merged this.

Nemikolh:
Yeah I agree, let's move forward with this PR 👍

PR #152, opened 2024-07-17
chore(style): minor code style adjustments for improved consistency

d3lm:
Small PR that updates some minor code styling to streamline consistency with the rest of the repo.

PR #154, opened 2024-07-18
chore: lint `*.astro` files

AriPerkkio:
- Include `*.astro` files in linting
- Fix prettier config

Note that target branch is base of #152, not `main`.


d3lm (approved):
Nice 💪

PR #165, opened 2024-07-24
feat: add `template.visibleFiles` option

AriPerkkio:
- Adds `template.visibleFiles` option that can be used to toggle files from template visible in lessons


AriPerkkio:
Running into https://github.com/stackblitz/tutorialkit/issues/167 here too. When "Solve" is clicked, files from template are shown as empty. Only files in `_solution` are shown correctly. 

Need to fix https://github.com/stackblitz/tutorialkit/issues/167 first.

AriPerkkio:
Tested manually with https://github.com/AriPerkkio/tutorial-vite-plugin/pull/15 - the amount of removed duplicate files is a lot!

As this changes quite a lot how editor's files are resolved, I think we should wait for @Nemikolh or @SamVerschueren to review this next week.

docs/tutorialkit.dev/src/content/docs/guides/creating-content.mdx
```diff
@@ -144,3 +158,75 @@ src/templates
...
+Editor's files are resolved in three steps. Each step overrides previous one:
+
+1. Display files matching `template.visibleFiles` (lowest priority)
+2. Display files from `_files` directory
+3. When solution is revealed, display files from `_solution` directory. (highest priority)
+
+<Tabs syncKey="file-visibilty">
+  <TabItem label="Initially">
+
+```markdown ins=/.{24}├── (first.js)/ ins=/└── (second.js)/ ins=/third.js/
```
AriPerkkio, line 173: This patterns might look a bit weird. It's matching the words on specific rows. I don't think expressive-code supports syntax like `ins={24, "first.js"}` to match `first.js` on line 24 only.

image

packages/astro/src/default/utils/content.ts
```diff
@@ -10,6 +10,7 @@ import type {
 import { folderPathToFilesRef, interpolateString } from '@tutorialkit/types';
 import { getCollection } from 'astro:content';
 import glob from 'fast-glob';
+import mm from 'micromatch';
```
Nemikolh, line 13: Super Nit pick: I would either import `isMatch` or call the import `micromatch`.

packages/astro/src/default/utils/content.ts
```diff
@@ -261,6 +264,22 @@ export async function getTutorial(): Promise<Tutorial> {
       ),
     };
 
+    if (lesson.data.template && typeof lesson.data.template !== 'string' && lesson.data.template.visibleFiles?.length) {
+      const templateFilesRef = await getFilesRefList(lesson.data.template.name, TEMPLATES_DIR);
```
Nemikolh, line 268:
```suggestion
      const [_, templateFiles] = await getFilesRefList(lesson.data.template.name, TEMPLATES_DIR);
```

packages/astro/src/default/utils/content.ts
```diff
@@ -261,6 +264,22 @@ export async function getTutorial(): Promise<Tutorial> {
...
+
+      for (const filename of templateFilesRef[1]) {
```
Nemikolh, line 270:
```suggestion
      for (const filename of templateFiles) {
```

packages/types/src/schemas/common.ts
```diff
@@ -170,8 +170,18 @@ export const webcontainerSchema = commandsSchema.extend({
       'Navigating to a lesson that specifies autoReload will always reload the preview. This is typically only needed if your server does not support HMR.',
     ),
   template: z
-    .string()
-    .optional()
+    .union([
+      // name of the template
+      z.string().optional(),
```
Nemikolh, line 175: Nit: I would move the `.optional()` to be after the `.union()`.

packages/types/src/schemas/common.ts
```diff
@@ -170,8 +170,18 @@ export const webcontainerSchema = commandsSchema.extend({
...
+
+      z.strictObject({
+        // name of the template
+        name: z.string(),
+
+        // list of globs of files that should be visible
+        visibleFiles: z.array(z.string()).optional(),
```
Nemikolh, line 182: A `describe` here would make the info for the vscode extension slightly better.

packages/astro/src/default/utils/content.ts
```diff
@@ -347,6 +366,15 @@ async function getFilesRefList(pathToFolder: string): Promise<FilesRefList> {
   return [filesRef, filePaths];
 }
 
+function formatTemplateFile(filename: string) {
+  // compare files without leading "/" so that patterns like ["src/index.js"] match "/src/index.js"
```
Nemikolh, line 370: I'm not super familiar with micromatch, but this comment and the logic after make me think that maybe this is not the library that we want?

I think we should have `visibleFiles` entries follow a similar format to `git`'s `gitignore` entries. That would lower the learning curve of TutorialKit a lot and not necessitate as much trial and error for users.

In particular if you say:

```yaml

template:
  visibleFiles:
    - /foobar
```

It should not match `/test/foobar` but match a `/foobar` file in the template.

If you mean to match every file named `foobar` then you can only say `foobar` or `**/foobar`.

packages/runtime/src/store/index.ts
```diff
@@ -353,3 +365,7 @@ export class TutorialStore {
     this._themeRef.set(this._themeRef.get() + 1);
   }
 }
+
+function filterEntries<T extends object>(obj: T, filter: string[]) {
+  return Object.fromEntries(Object.entries(obj).filter(([entry]) => filter.includes(entry)));
+}
```
Nemikolh, line 371: That seems a bit overkill, I would use a much simpler function for what we're doing here because we already know that each value in `filter` is expected to be an entry in `obj`. 

Let's rename this function `pick` and do:

```ts
function pick<T>(obj: Record<string, T>, entries: string[]): Record<string, T> {
  const result = {};

  for (const entry of entries) {
    // kept to avoid having undefined values, but really if this is hit this is a
    // bug so maybe we should log something if we have an entry that is not present
    if (entry in obj) {
      result[entry] = obj[entry];
    }
  }

  return result;
}
```

Nemikolh (commented):
Looks good! I have a couple of comments but my biggest point is that I would want us to have `visibleFiles` entries match's git's `.gitignore` format.

On a TutorialKit project, I expect most of our users to use git to version their tutorial and so I think it's reasonable to expect from them to be more familiar with the `.gitignore` format than something else. I think it's similar to globs but a comment in this PR make me think that maybe it doesn't work exactly the same.

packages/runtime/src/store/index.ts
AriPerkkio: Looks better,  nice!

packages/astro/src/default/utils/content.ts
AriPerkkio: Sure! This is how we've used it in Vitest: https://github.com/search?q=repo%3Avitest-dev%2Fvitest%20micromatch&type=code
Though it looks like in Vite it's imported as `micromatch`. The package is CJS and doesn't seem to work with named imports.

packages/astro/src/default/utils/content.ts
AriPerkkio: I picked `micromatch` since it's already a transitive dependency from `fast-glob`.

`.gitignore` doesn't require prefixing all files with `/`? Meaning that:

```sh
$ touch new-file.ts
$ git status

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        new-file.ts

$ echo "new-file.ts" >> .gitignore 
$ git status

Changes not staged for commit:
        modified:   .gitignore
```

What `formatTemplateFile` here does, is that it allows you to defined `visibleFile: ["src/main.js"]` instead of `["/src/main.js"]`. Do we want to revert this `/` handling?

AriPerkkio: Though I guess metadata's `focus` option does require `/` prefixes for all files.

AriPerkkio: @Nemikolh here's small playground to check how it works:
https://stackblitz.com/~/edit/stackblitz-starters-efozpk?file=index.mjs
`node index.mjs` to run.

Nemikolh: I modified the example a bit to showcase what I mean: https://stackblitz.com/~/edit/stackblitz-starters-bq7pxo.

So basically when we have:

```js
[
  '/foobar/a.js',
  '/a.js',
]
```

If you have a `.gitignore` that says:

```
a.js
```

then it will ignore both files. But if it says `/a.js`, it will only ignore the second one.

Nemikolh: Oh looks like this is working! This might be good enough?

Nemikolh: cc @SamVerschueren

AriPerkkio: Now using `basename: true`.

packages/runtime/src/store/index.ts
```diff
@@ -165,15 +173,17 @@ export class TutorialStore {
...
 
-        this._lessonTemplate = template;
         this._lessonFiles = files;
         this._lessonSolution = solution;
+        this._lessonTemplate = template;
+        this._visibleTemplateFiles = pick(template, lesson.files[1]);
 
-        this._editorStore.setDocuments(files);
+        const editorFiles = { ...this._visibleTemplateFiles, ...this._lessonFiles };
+        this._editorStore.setDocuments(editorFiles);
```
Nemikolh, line 182: I just noticed that the files from the template are not read-only.

Given that they aren't part of the lesson but are here to provide information on the project, I really feel like they should be read only. Otherwise it feels like we're moving too far away from the original goal of having tutorials be this "safe" environment where learners can focus on the learning.

Nemikolh (commented):
Noticed a really important detail that regardless of the behaviour of the matching should be addressed.

packages/runtime/src/store/index.ts
AriPerkkio: I don't think we have options to make any file readonly at the moment. Maybe we need such feature, especially ones adding new files via terminal becomes possible?

Currently it's intentional that files are modifiable. This is simply to reduce repetition - you don't have to copy-paste files in every single lesson just to make them visible.

Nemikolh: > I don't think we have options to make any file readonly at the moment. Maybe we need such feature, especially once adding new files via terminal becomes possible?

That is correct, until this PR we didn't have a need for them.

I'm not quite sure what we want to do about files being added from a terminal or more generally added in WebContainer by any mean, but I'm tempted to think that we'll want to design a solution that around a specific example.

> Currently it's intentional that files are modifiable. This is simply to reduce repetition - you don't have to copy-paste files in every single lesson just to make them visible.

Then IMO it's the wrong abstraction. I never saw `visibleFiles` as intended to reduce repetition in `_files`. If we want that then we should maybe consider allowing `.tk-config.json` in `_files` (which I think already works but isn't documented) or have a different approach. This feels a bit hacky and against the original idea that templates are not part of the lesson (as in initial files to modify or the final solution).

AriPerkkio: > Otherwise it feels like we're moving too far away from the original goal of having tutorials be this "safe" environment where learners can focus on the learning.

Yep, it sounds like we need to reconsider these changes. Maybe it's best to have only files from `_files` visible, as it currently is. Having files that aren't modifiable in file tree could make tutorials even more confusing.

Nemikolh: Oh that's a good point actually! It's not really a template property but more like what we should show from the fs regardless of where it's coming from.

Hmm I agree maybe this is something we should discuss further, maybe we could start a thread on twitter about it and ping people that have requested the feature? Or maybe we can have the conversation in a GitHub issue.

docs/tutorialkit.dev/src/content/docs/guides/creating-content.mdx
AriPerkkio: Missing `_files/src` here 😱

packages/runtime/src/store/index.ts
AriPerkkio: Maybe something like:

```
editor: 
  fileTree: 
     include: ["..."]
     templateFiles: ["..."]
```

PR #170, opened 2024-07-25
fix(docs): jumping to bottom of getting started page

EricSimons:


d3lm (approved):
Fixed the title so it matches our convention. LGTM otherwise 👍

AriPerkkio (approved):
Good catch! 💯 

This doesn't come up easily on desktops as whole page height can fit the viewport easily.

Issue #189, opened 2024-07-30
`templates/*/node_modules` should be ignored


AriPerkkio:
### Describe the bug

If template's `node_modules` is present, these are copied to webcontainer. This can easily lead into copying 10s of megabytes of content. Also platform specific modules will not work as they are not using wasm builds of native modules.

- https://github.com/stackblitz/tutorialkit/discussions/187
- https://github.com/stackblitz/tutorialkit/issues/190
- https://github.com/stackblitz/tutorialkit/issues/196

### Steps to reproduce

1. `npm create tutorial`
2. `cd src/templates/default; npm install`
3. Start dev server
4. Preview fails to load

### Expected behavior

`node_modules` should be excluded by default. 

### Screenshots

<img src="https://github.com/user-attachments/assets/5570f702-92c8-4ea9-b40c-d8a529331a3b" width="400" />
<br />
<img src="https://github.com/user-attachments/assets/c7c5d108-c274-4f67-a0b4-e806ec2503db" width="400" />

### Platform

- TutorialKit version: 0.1.1

### Additional

https://github.com/stackblitz/tutorialkit/blob/d21d5e9062687922b8eb8d43e57611887f287cbf/packages/astro/src/webcontainer-files/constants.ts#L1

AriPerkkio:
Though it's completely valid use case to add pre-optimized `node_modules` into templates to make tutorials load faster. 

To get the best user experience here, we should ignore `templates/*/node_modules` by default. Including `templates/*/node_modules` should be possible by opt-in option. Maybe after https://github.com/stackblitz/tutorialkit/pull/165 we could have:

```yaml
template:
  name: my-template
  ignoredFiles: [] # Defaults to '**/node_modules**/', [] makes node_modules included
```

AriPerkkio:
So far I've tried two approaches:
- Add `template.ignoredFiles` into metadata
  - This way we would still end up creating those 50MB `<name>-template.json` files, as exclusion would happen after the files have been loaded to the browser
- Add `ignoredFiles` to `.tk-config` file, that is already used to configure `src/templates`
  - This would work well, but now it's not possible to set this in tutorial/part/chapter/lesson metadata
  - Implementing this into codebase is also a bit difficult now, but possible

More users are now running into this so we should publish a release with quick fix as soon as possible. I think I'll make `node_modules` exclusion non-configurable now, and create a separate feature issue for configuring `ignoredFile` in metadata or `.tk-config.json`.

[...]

PR #198, opened 2024-08-01
fix: ignore templates `node_modules`

AriPerkkio:
- Fixes https://github.com/stackblitz/tutorialkit/issues/189

Fixes issues where users would see `jsh: permission denied` errors if `src/templates/*/node_modules` were present on host machines' file system. TutorialKit would copy those `node_modules` to webcontainers and try to run them. These `node_modules` may contain platform specific binaries that fail to run on webcontainers (non-wasm builds). 

Adds `node_modules` to be ignored in `src/templates`. At this point the change is hard-coded and not configurable. See https://github.com/stackblitz/tutorialkit/issues/189#issuecomment-2262036287.

I'll create a follow-up feature request that would allow including `node_modules`, so that tutorial authors could include pre-optimized dependencies in their templates. This is lower priority task compared to #189 that users are running into now.


d3lm (approved):
LGTM

Issue #211, opened 2024-08-04
Add (way to add) Svelte language support to CodeMirror editor


adrianmroz:
### Is your feature request related to a problem?

I want to create tutorial for Svelte using tutotiralkit, but provided CodeMirror instance does not support Svelte files ie. no syntax highlighting.

### Describe the solution you'd like.

Add Svelte language description to CodeMirror inside [CodeMirrorEditor/languages.ts](https://github.com/stackblitz/tutorialkit/blob/main/packages/components/react/src/core/CodeMirrorEditor/languages.ts)

### Describe alternatives you've considered.

Good alternative is to give developer a way to customize CodeMirror instance. Would be better in the long term. On the other hand, you could take a shortcut and add Svelte support directly. Your blogpost mentions Svelte a lot and I imagine community would benefit even from such shortcut.

### Additional context

I found this package: [@replit/codemirror-lang-svelte](https://github.com/replit/codemirror-lang-svelte) which is also used by code mirror instance on [learn.svelte.dev](https://learn.svelte.dev/tutorial/welcome-to-svelte)

PR #212, opened 2024-08-04
feat: add Svelte language to CodeMirror

adrianmroz:
Should solve #211


Nemikolh (commented):
Hey @adrianmroz! Thanks a lot for the PR! 😃 

This looks pretty good! 

Only a few more steps and we can merge your PR:

 1. The dependency seems to need a few `peerDependencies`: `@lezer/common`, `@lezer/javascript`, `@lezer/lr`. Could you add them explicitly to the `package.json` as well?
 2. The lock file needs to be updated. You can do that with `pnpm i` at the root of the workspace.

If you want to test your changes you can then do:

```
pnpm dev
```

You can add a temporary svelte file in `packages/template/src/content/tutorial/1-basics/1-introduction/1-welcome/_files`, you navigate to http://localhost:4321/ and check that the syntax highlighting works as expected.

Let me know if you run into any issues!

adrianmroz:
@Nemikolh 

I've pushed changes to dependencies and lock file.

I also checked example file for syntax highlighting. It is not perfect, but much better than nothing :) 
image

Let me know if I still missed something.

Nemikolh (approved):
Looks good! Thanks for the PR 😃

Issue #193, opened 2024-07-30
Add "Open in StackBlitz" button


sulco:
### Is your feature request related to a problem?

Currently, there's no way for a student to see (or run) the whole app, or persist the changes. Adding "Open in StackBlitz" will allow to continue playing with the codebase of a specific lesson.

It will also be useful for creating reproductions for broken lessons.

### Describe the solution you'd like.

Let's add the "Open in StackBlitz" button to the header, like this: 
![3](https://github.com/user-attachments/assets/7ac84322-44f3-4888-9700-8d15d05f2bba)

### Additional context

This can be achieved with the [StackBlitz Post API](https://developer.stackblitz.com/platform/api/post-api)

PR #219, opened 2024-08-06
feat: add 'Open in StackBlitz'-button

AriPerkkio:
- Closes https://github.com/stackblitz/tutorialkit/issues/193

Adds an button on top bar that can be used to open current lesson and template code in separate StackBlitz project. Current code from editor is applied on top of the files.

Users can disable showing this button by defining `openInStackBlitzLink: false` in frontmatter metada. By default it's always visible.


packages/astro/src/default/components/OpenInStackblitzLink.astro
```diff
@@ -0,0 +1,93 @@
...
+    <path fill="currentColor" d="M12.747 16.273h-7.46L18.925 1.5l-3.671 10.227h7.46L9.075 26.5l3.671-10.227z"></path>
+  </svg>
+</button>
+
+<script>
+  import StackBlitzSDK from '@stackblitz/sdk';
+  import { tutorialStore } from '../components/webcontainer.js';
+
+  // initialize handlers on each page load as it's possible some pages disable openInStackBlitzLink
+  document.addEventListener('astro:page-load', onInit);
```
AriPerkkio, line 17: This is required even when `transition:persist` is used due to:

- Disable `openInStackBlitzLink` on a single lesson
- Open that lesson in fresh browser tab. The element is not run.
- Navigate to a page that doesn't have `openInStackBlitzLink` disabled. Script tags are not run, handler is not attached.

packages/astro/src/default/components/OpenInStackblitzLink.astro
```diff
@@ -0,0 +1,93 @@
...
+    const files: Record<string, string> = {};
+
+    // first add template files
+    for (const [filePath, value] of Object.entries(tutorialStore.template || {})) {
+      files[removeLeadingSlash(filePath)] = value.toString();
+    }
+
+    // next overwrite with files from editor
+    for (const { filePath, value } of Object.values(tutorialStore.documents.get())) {
+      files[removeLeadingSlash(filePath)] = value.toString();
```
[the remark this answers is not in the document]
Nemikolh: Images mostly but any non-text data that users could endup having in WebContainer. I think you should skip elements for now. I don't know if we have a way so that they're processed correctly on StackBlitz atm.

Nemikolh:
Really cool PR! A few things we should do:
* Move the logic of taking the snapshot that get sent to StackBlitz in a `takeSnapshot` function on the `TutorialRunner` class. Then the tutorialStore could also expose the same function calling the underlying `_tutorialRunner`.
* Skip non-text files for now. I don't know if this is supported by the SDK. We can always revisit that later.

The rest is really great! I really love the idea of generating a "start" command in the `package.json` that does the same as the current lesson. The logo is perfect, I love it ❤️

AriPerkkio:
Added `takeSnapshot` to runner that prepares `files`. The lesson data wasn't available there so `title` and `description` are resolved in button component as before. Let me know what you think!

packages/astro/src/default/components/OpenInStackblitzLink.astro
```diff
@@ -23,71 +23,17 @@
...
-    if (files['package.json']) {
-      const mainCommand = resolveCommand(lesson.data.mainCommand);
-      const prepareCommands = (lesson.data.prepareCommands || []).map(resolveCommand);
-      const startCommand = [...prepareCommands, mainCommand].filter(Boolean).join(' && ');
-
-      files['package.json'] = JSON.stringify({ ...packageJson, stackblitz: { startCommand } }, null, 2);
-    }
-
     StackBlitzSDK.openProject({
       title: lesson.data.title || 'Lesson',
```
Nemikolh, line 33: I'm starting to think that maybe this should instead use the title of the tutorial instead of the title of the lesson.

Even if that means the same name for every lesson I think this will avoid strange names.

What do you think?

packages/runtime/src/tutorial-runner.ts
```diff
@@ -302,6 +302,43 @@ export class TutorialRunner {
...
+    for (const [filePath, value] of Object.entries(this._currentFiles || {})) {
+      if (typeof value === 'string') {
+        files[removeLeadingSlash(filePath)] = value;
+      }
+    }
+
+    const packageJson = parseJson(files['package.json']);
+
+    // add start commands
+    if (files['package.json']) {
```
Nemikolh, line 331: This should check if there is already a start or dev script because then we don't need to do anything. StackBlitz default logic will be good enough I think.

packages/runtime/src/tutorial-runner.ts
```diff
@@ -302,6 +302,43 @@ export class TutorialRunner {
...
+    }
+
+    // next overwrite with files from editor
+    for (const [filePath, value] of Object.entries(this._currentFiles || {})) {
+      if (typeof value === 'string') {
+        files[removeLeadingSlash(filePath)] = value;
+      }
+    }
+
+    const packageJson = parseJson(files['package.json']);
```
Nemikolh, line 328: We should probably reuse `this._packageJsonContent`? I just realised that the `package.json`, if present, might be in a sub folder.

Note that the path, if it's at the root, will be `/package.json` instead of `package.json`.

packages/runtime/src/tutorial-runner.ts
```diff
@@ -484,3 +521,19 @@ async function updateFiles(webcontainer: WebContainer, previousFiles: Files, new
 
   await webcontainer.mount(toFileTree(addedOrModified));
 }
+
+function removeLeadingSlash(filePath: string) {
+  if (filePath.startsWith('/')) {
```
Nemikolh, line 526: This is always true where the function is used. I would probably remove the function and inline the slice.

packages/runtime/src/tutorial-runner.ts
```diff
@@ -484,3 +521,19 @@ async function updateFiles(webcontainer: WebContainer, previousFiles: Files, new
...
+    return filePath.slice(1);
+  }
+
+  return filePath;
+}
+
+function parseJson(deserialized: undefined | string): any {
+  try {
+    return JSON.parse(deserialized || '{}');
```
Nemikolh, line 535: I find it surprising that parsing undefined returns an empty object.

In the code above where this function is used it would be more correct to leave the `package.json` unmodified if it isn't valid JSON. Otherwise this code will throw away everything else.

Maybe this auxiliary function is too generic. I think it's probably better to inline the logic.

Nemikolh (commented):
I like where this is going! Still a few things to fix and it should be good to go 🙌

packages/astro/src/default/components/OpenInStackblitzLink.astro
AriPerkkio: What's title of the tutorial? Setting `title` in frontmatter for `type: tutorial` is not valid. Following configuration will throw validation error:

```md
---
type: tutorial
title: Example
---
```

packages/runtime/src/tutorial-runner.ts
AriPerkkio: The `_packageJsonContent` doesn't contain contents of `package.json` if that file is not present in `_files`. We want to read `package.json`'s from templates too. 

Should I make changes to pass `{...files, ...template}` here?

https://github.com/stackblitz/tutorialkit/blob/298615748b1f2d3ea737c591ce193eb0d28407ca/packages/runtime/src/tutorial-runner.ts#L191

It's then set here:

https://github.com/stackblitz/tutorialkit/blob/298615748b1f2d3ea737c591ce193eb0d28407ca/packages/runtime/src/tutorial-runner.ts#L416-L420

Nemikolh: Oh good catch! That's indeed a bug! Yep, line 191 should definitely see `{...files, ...template}`.

packages/astro/src/default/components/OpenInStackblitzLink.astro
Nemikolh: So the idea is to add `title` as an optional property of `tutorialSchema` in `types/src/schemas/tutorial.ts`. That way we don't break existing users and don't force them to have one.

For the default value here I would set something like `Project generated by TutorialKit` instead of `Lesson`.

packages/runtime/src/tutorial-runner.ts
```diff
@@ -314,26 +314,32 @@ export class TutorialRunner {
...
-      const prepareCommands = (this._currentRunCommands?.prepareCommands || []).map((c) => c.shellCommand);
-      const startCommand = [...prepareCommands, mainCommand].filter(Boolean).join(' && ');
+      let packageJson;
+
+      try {
+        packageJson = JSON.parse(files['package.json']);
+      } catch {}
+
+      // add start commands when missing
+      if (packageJson && !packageJson.stackblitz?.startCommand) {
```
Nemikolh, line 336:
```suggestion
      if (packageJson && !packageJson.scripts?.start && !packageJson.scripts?.dev) {
```

packages/astro/src/default/components/OpenInStackblitzLink.astro
AriPerkkio: > So the idea is to add title as an optional property of `tutorialSchema` in `types/src/schemas/tutorial.ts`.

Do we want to show this somewhere on the UI as well? Not sure if this is getting out-of-scope regarding this PR. 🤔

packages/runtime/src/tutorial-runner.ts
AriPerkkio: Sure! Though I wonder if tutorial's `prepareCommands` could have something that project's `start` and `dev` scripts don't cover. 🤔

packages/runtime/src/tutorial-runner.ts
Nemikolh: We also need to do add a:

```
this._packageJsonPath = filePath;
```

To know where the file is and reuse that when modifying it.

packages/runtime/src/tutorial-runner.ts
Nemikolh: Yeah that's a good point, let's keep it like you have done. We can always revisit that part later.

packages/astro/src/default/components/OpenInStackblitzLink.astro
Nemikolh: > Do we want to show this somewhere on the UI as well?

I don't think we want to. But now that you've said that I don't think my proposition makes much sense actually.

What if instead we used a different name of the property `openInStackBlitzLink` to be `openInStackBlitz`:

```ts
openInStackBlitz: z.union([
  z.boolean(), // to have it allowed or disallowed
  z.strictObject({
    projectTitle: z.string().optional(),
    projectDescription: z.string().optional(),
  }),
])
```

I'm also fine with not using anything and hardcoding `Project generated by TutorialKit` for now. I just think that people will probably want to tweak this at some point, so if we could add it here the better.

AriPerkkio: The object-type config is great idea! I've implemented that now. In future we might want to expose options like `template` there, as we currently hardcode `template: 'node'` there.

Nemikolh: Oh that's a really good point!! 🤩  I feel like we should add it too, what do you think? 😃

docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx
```diff
@@ -211,6 +211,15 @@ You can instruct Github to show the source code instead by adding `plain=1` quer
...
-### `openInStackBlitzLink`
+### `openInStackBlitz`
 Display a link for opening current lesson in StackBlitz.
-<PropertyTable inherited type="boolean" />
+<PropertyTable inherited type="OpenInStackBlitz" />
+
+The `OpenInStackBlitz` type has the following shape:
+
+```ts
+type OpenInStackBlitz = string
```
Nemikolh, line 221:
```suggestion
type OpenInStackBlitz =
```
(string is not a valid value)

Nemikolh (approved):
YAY!! Love this! Really good work on this front! 🤩 🥳

packages/astro/src/default/components/OpenInStackblitzLink.astro
AriPerkkio: Added `projectTemplate` now.

Nemikolh (approved):
Good job on this! 🤩

PR #223, opened 2024-08-07
feat(extension): support config-based ordering

Nemikolh:
This PR adds:

 * ✅ Support for config based ordering.
 * ✅  Deletion of lessons, chapters or parts (I needed that for testing).
 * ✅ It now shows the root node so that we can view the tutorial config. The name of the root node is the folder name of the project.

It does not implement re-ordering via drag and drop (I'll do that in a follow up PR).

I simplified the TreeView so that it does not contains any logic regarding the data structure generation. This is now done entirely in `models/tree/load.ts` (not a fan of this location but couldn't think of a better one). The logic no longer use `fs.readdirSync` and instead use `vscode.workspace.fs` functions.

https://github.com/user-attachments/assets/a860af2f-c9d8-4fe1-9066-8a1a128b02dc

### TODO before merging

 * [ ] Respect config based ordering when creating a new unit


extensions/vscode/package.json
```diff
@@ -119,10 +138,6 @@
     ]
   },
   "scripts": {
-    "__esbuild-base": "esbuild ./src/extension.ts --bundle --outfile=dist/extension.js --external:vscode --format=cjs --platform=node",
-    "__dev": "pnpm run esbuild-base -- --sourcemap --watch",
-    "__vscode:prepublish": "pnpm run esbuild-base -- --minify",
-    "__build": "vsce package",
```
Nemikolh, line 125: I removed those because I don't think we should use them? @sulco what do you think?

extensions/vscode/src/views/lessonsTree.ts
```diff
@@ -1,114 +1,98 @@
...
-        if (metadataFile) {
-          const metadataFilePath = path.join(filePath, metadataFile);
-          const metadataFileContent = fs.readFileSync(metadataFilePath, 'utf8');
-          const parsedContent = grayMatter(metadataFileContent);
+      clearTimeout(timeoutId);
 
-          lesson.name = parsedContent.data.title;
+      timeoutId = setTimeout(async () => {
+        loading = true;
+        await this.refresh();
```
Nemikolh, line 37: When a file is changed, we throw away everything and recalculate everything (lazily on a needed basis though). 

This is pretty inefficient but very simple.

Ideally, I'd want to do something more fine-grained where we only update the element of the tree that need to. This should improve perf on really big tutorials with many many markdown files.

extensions/vscode/src/commands/tutorialkit.delete.ts
```diff
@@ -0,0 +1,18 @@
...
+  if (nodes.length === 0) {
+    nodes = getLessonsTreeView().selection;
+  }
+
+  for (const node of nodes) {
+    await vscode.workspace.fs.delete(node.path, { recursive: true });
+  }
+
+  return cmd.refresh();
+}
```
AriPerkkio, line 18: Should this also delete the unit from metadata? Right now when you delete a lesson it is still present in custom ordered chapter metadata. 

For example, remove `One` from the tree view in this case. Should the `lessons` be modified here as well?

```md
---
type: chapter
title: Example
lessons:
  - One
  - Two
---
```

extensions/vscode/src/utils/uri.ts
```diff
@@ -0,0 +1,16 @@
...
+import { dirname } from 'path';
+
+export function uriDirname(uri: vscode.Uri): vscode.Uri {
+  if (uri.path.length === 0 || uri.path === '/') {
+    return uri;
+  }
+
+  let path = dirname(uri.path);
+
+  if (path.length === 1 && path.charCodeAt(0) === 46 /* CharCode.Period */) {
```
AriPerkkio, line 11: Why not just `if (path === ".")` ?

AriPerkkio (commented):
I'm not that familiar with `vscode` extension API but the code itself looks good.

extensions/vscode/src/commands/tutorialkit.delete.ts
Nemikolh: Oh really good catch! 😍 Will make sure the parent is updated 👍

extensions/vscode/src/utils/uri.ts
Nemikolh: Oh good point! I copied that from https://www.npmjs.com/package/vscode-uri to test it but forgot to remove it. I'll add a dependency to `vscode-uri` and use the function from that package directly instead.

AriPerkkio (approved):
Looks good to me. I did some testing locally with Vite Plugin Tutorial too. Nice work! 💯

PR #224, opened 2024-08-07
docs: add info on how to create callouts

henrikvilhelmberglund:
This PR adds info on how to actually create the callouts which are shown in the `theming` reference. I don't usually use Astro so I was a bit confused when the "normal" Github style of `> [!NOTE] ` etc didn't work and I had to google around a bit to find how to actually create them.

Maybe it doesn't really fit in the theming page but when you search for `callouts` you get to the theming page so maybe it's fine.


Nemikolh (approved):
@henrikvilhelmberglund Thanks a lot for the PR! 😍  This sounds pretty reasonable to me!

AriPerkkio (approved):

PR #229, opened 2024-08-08
chore: release CLI v0.1.4

stackblitz-gitops:
Bump CLI to version 0.1.4.

PR #232, opened 2024-08-08
build: use node lts

AriPerkkio:
Should help with issues like #229


AriPerkkio:
Where is the cloudflare bot's automatic deploy... 🤷

Issue #192, opened 2024-07-30
Adding urls to the `previews` definition


noam-honig:
### Is your feature request related to a problem?

I want to show several preview tabs - with different relative urls for example:
{port:5173, title:"frontend"}
{port:5173, title:"api",url:"/api/tasks"}
{port:5173, title:"admin",url:"/api/admin"}

### Describe the solution you'd like.

I want to show several preview tabs - with different relative urls for example:
{port:5173, title:"frontend"}
{port:5173, title:"api",url:"/api/tasks"}
{port:5173, title:"admin",url:"/api/admin"}

### Describe alternatives you've considered.

reviewed the https://tutorialkit.dev/reference/configuration/ url and couldn't find anything

AriPerkkio:
Currently this is not possible but we are planning to add support for it. I've done some design for the possible API earlier:

> ```yaml
> previews: 
>    - 1234/pathname
>    - [1234/pathname, 'Title']
>    - { port: 1234, path: '/pathname', title: 'Title' }
> ```

AriPerkkio:
Released in `0.1.5`.

PR #233, opened 2024-08-09
feat(runtime): add `preview.pathname`

AriPerkkio:
- Closes https://github.com/stackblitz/tutorialkit/issues/192

Adds support for defining pathname for previews. Example from Vite Plugin Tutorial where I'd like to show two previews for same server:

```yaml
---
type: lesson
title: Extra - Source map support
focus: /vite.config.ts
previews:
  - { port: 5173, title: "Vite" }
  - { port: 5173, title: "Inspect", pathname: "/__inspect/" }
---
```

This will also be useful for remult tutorial: https://learn.remult.dev/1-basics/5-live-query/2-realtime-updates - mentioned here https://github.com/stackblitz/tutorialkit/discussions/226

### DIFF
--- a/docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx
+++ b/docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx
@@ -91,16 +91,30 @@ Configure whether or not the editor should be rendered. If an object is provided
 ##### `previews`
 Configure which ports should be used for the previews allowing you to align the behavior with your demo application's dev server setup. If not specified, the lowest port will be used.
 
-You can optionally provide these as an array of tuples where the first element is the port number and the second is the title of the preview, or as an object.
 <PropertyTable inherited type={'Preview[]'} />
 
 The `Preview` type has the following shape:
 
 ```ts
-type Preview = string
+type Preview =
+    | number
+    | string
     | [port: number, title: string]
-    | { port: number, title: string }
+    | [port: number, title: string, pathname: string]
+    | { port: number, title: string, pathname?: string }
 
+```
+
+Example value:
+
+```yaml
+previews:
+  - 3000 # Preview is on :3000/
+  - "3001/docs" # Preview is on :3001/docs/
+  - [3002, "Dev Server"] # Preview is on :3002/. Displayed title is "Dev Server".
+  - [3003, "Dev Server", "/docs"] # Preview is on :3003/docs/. Displayed title is "Dev Server".
+  - { port: 3004, title: "Dev Server" } # Preview is on :3004/. Displayed title is "Dev Server".
+  - { port: 3005, title: "Dev Server", pathname: "/docs" } # Preview is on :3005/docs/. Displayed title is "Dev Server".
 ```
 
 ##### `mainCommand`
--- a/packages/runtime/src/store/previews.ts
+++ b/packages/runtime/src/store/previews.ts
@@ -4,7 +4,7 @@ import { PreviewInfo } from '../webcontainer/preview-info.js';
 import type { WebContainer } from '@webcontainer/api';
 
 export class PreviewsStore {
-  private _availablePreviews = new Map<number, PreviewInfo>();
+  private _availablePreviews: PreviewInfo[] = [];
   private _previewsLayout: PreviewInfo[] = [];
 
   /**
@@ -21,18 +21,21 @@ export class PreviewsStore {
     const webcontainer = await webcontainerPromise;
 
     webcontainer.on('port', (port, type, url) => {
-      let previewInfo = this._availablePreviews.get(port);
+      const previewInfos = this._availablePreviews.filter((preview) => preview.port === port);
 
-      if (!previewInfo) {
-        previewInfo = new PreviewInfo(port, type === 'open');
-        this._availablePreviews.set(port, previewInfo);
+      if (previewInfos.length === 0) {
+        const info = new PreviewInfo(port, type === 'open');
+        previewInfos.push(info);
+        this._availablePreviews.push(info);
       }
 
-      previewInfo.ready = type === 'open';
-      previewInfo.baseUrl = url;
+      previewInfos.forEach((info) => {
+        info.ready = type === 'open';
+        info.baseUrl = url;
+      });
 
       if (this._previewsLayout.length === 0) {
-        this.previews.set([previewInfo]);
+        this.previews.set(previewInfos);
       } else {
         this._previewsLayout = [...this._previewsLayout];
         this.previews.set(this._previewsLayout);
@@ -58,14 +61,12 @@ export class PreviewsStore {
     const previewInfos = previews.map((preview) => {
       const info = new PreviewInfo(preview);
 
-      let previewInfo = this._availablePreviews.get(info.port);
+      let previewInfo = this._availablePreviews.find((availablePreview) => PreviewInfo.equals(info, availablePreview));
 
       if (!previewInfo) {
         previewInfo = info;
 
-        this._availablePreviews.set(previewInfo.port, previewInfo);
-      } else {
-        previewInfo.title = info.title;
+        this._availablePreviews.push(previewInfo);
       }
 
       return previewInfo;
--- a/packages/runtime/src/store/previews.ts
+++ b/packages/runtime/src/store/previews.ts
@@ -34,7 +34,7 @@ export class PreviewsStore {
       portInfo.origin = url;
 
       if (this._previewsLayout.length === 0) {
-        this.previews.set([new PreviewInfo(portInfo)]);
+        this.previews.set([new PreviewInfo({}, portInfo)]);
       } else {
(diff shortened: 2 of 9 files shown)


packages/runtime/src/store/previews.ts
```diff
@@ -58,14 +61,12 @@ export class PreviewsStore {
...
 
-      let previewInfo = this._availablePreviews.get(info.port);
+      let previewInfo = this._availablePreviews.find((availablePreview) => PreviewInfo.equals(info, availablePreview));
 
       if (!previewInfo) {
         previewInfo = info;
 
-        this._availablePreviews.set(previewInfo.port, previewInfo);
-      } else {
-        previewInfo.title = info.title;
```
AriPerkkio, line 68: Any ideas why this was here? Removing it doesn't seem to break anything.

docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx
```diff
@@ -91,18 +91,32 @@ Configure whether or not the editor should be rendered. If an object is provided
...
+Example value:
+
+```yaml
+previews:
+  - 3000 # Preview is on :3000/
+  - "3001/docs" # Preview is on :3001/docs/
+  - [3002, "Dev Server"] # Preview is on :3002/. Displayed title is "Dev Server".
+  - [3003, "Dev Server", "/docs"] # Preview is on :3003/docs/. Displayed title is "Dev Server".
+  - { port: 3004, title: "Dev Server" } # Preview is on :3004/. Displayed title is "Dev Server".
+  - { port: 3005, title: "Dev Server", pathname: "/docs" } # Preview is on :3005/docs/. Displayed title is "Dev Server".
```
Nemikolh, line 117: Oh this is really nice!

packages/runtime/src/store/previews.ts
```diff
@@ -4,7 +4,7 @@ import { PreviewInfo } from '../webcontainer/preview-info.js';
 import type { WebContainer } from '@webcontainer/api';
 
 export class PreviewsStore {
-  private _availablePreviews = new Map<number, PreviewInfo>();
+  private _availablePreviews: PreviewInfo[] = [];
```
Nemikolh, line 7: Oh interesting. I would prefer if we fixed it using a `PortInfo` class instead.

packages/runtime/src/store/previews.ts
```diff
@@ -21,18 +21,21 @@ export class PreviewsStore {
...
 
-      previewInfo.ready = type === 'open';
-      previewInfo.baseUrl = url;
+      previewInfos.forEach((info) => {
+        info.ready = type === 'open';
+        info.baseUrl = url;
+      });
 
       if (this._previewsLayout.length === 0) {
-        this.previews.set([previewInfo]);
```
Nemikolh, line 35: This will become:

```ts
this.previews.set(new PreviewInfo(portInfo));
```

packages/runtime/src/store/previews.ts
Nemikolh: This is because the preview title can be changed between lessons even if the port hasn't changed.

However now, if we change `_availablePreviews` to be `Map<number, PortInfo>` and instead have all `PreviewInfo` have a reference to an instance in that map, then in here we always re-create the `PreviewInfo` instances.

Nemikolh (commented):
I like the config change!

I would prefer if we had `_availablePreviews` set to `Map<number, PortInfo>` instead which it should have been from the beginning. The idea behind `_availablePreviews` was to track information about the active `port`s from a webcontainer perspective.

However, when I had written the code I thought that most use case would probably have a single preview per port which is an over-simplification.

The correct approach is to use a different class in `_availablePreviews` other than `PreviewInfo` and make sure all `PreviewInfo` instances point to that instance.

So that means:

```ts
class PortInfo {
   constructor(
     public readonly port: number,
     public origin: string;
     public ready: boolean
  ) {}
}
```

Then we modify `PreviewInfo` to something like:

```diff
export class PreviewInfo {
-  port: number;
-  ready: boolean;
-  baseUrl?: string;
+  readonly portInfo: PortInfo; // initialized in constructor
+
+  get port() { return this.portInfo.port; }
+  get baseUrl() { return this.portInfo.origin; }
```

Finally when `setPreviews` is called we check if a `PortInfo` is present in the `Map`. If it isn't then we create one, and in `webcontainer.on('port')` we update that instance instead.

We could also consider have the `ready` and `origin` fields in `PortInfo` be `atom`s which I think could make sense, however given that in the current code we signal changes on previews by setting the `previews` `atom` on `PreviewsStore` to a new array instance, we don't really need those to be `atom` I think.

AriPerkkio:
> I would prefer if we had `_availablePreviews` set to `Map<number, PortInfo>`

> ```diff
> export class PreviewInfo {
> +  readonly portInfo: PortInfo; // initialized in constructor
> ```

I think we need to either make it `Map<number, PortInfo[]>` or public `portInfo`, so that we can mark multiple `PortInfo`'s ready at once.

If it was public field that allows overwriting, we would do it here:
```ts
    const previewInfos = previews.map((preview) => {
      const info = new PreviewInfo(preview);
      const portInfo = this._availablePreviews.get(info.port);

      if (!portInfo) {
        this._availablePreviews.set(info.port, info.portInfo);
      } else {
        info.portInfo = portInfo; // <-- All previews sharing same port should point to same portInfo
      }

      return info;
    });
```

Or if we would track multiple `PortInfo[]` in the map, we could then mark all of them ready on webcontainer's `port` event.

packages/runtime/src/store/previews.ts
AriPerkkio: Applied this, though this required `PreviewInfo` to accept `PortInfo` as constructor argument.

packages/runtime/src/store/previews.ts
```diff
@@ -57,18 +59,15 @@ export class PreviewsStore {
...
-
-      if (!previewInfo) {
-        previewInfo = info;
-
-        this._availablePreviews.set(previewInfo.port, previewInfo);
+      if (!portInfo) {
+        this._availablePreviews.set(info.port, info.portInfo);
       } else {
-        previewInfo.title = info.title;
+        info.portInfo = portInfo;
```
AriPerkkio, line 67: When webcontainer's `port` event is emitted, we mark a single `PortInfo` as ready. This single `PortInfo` can be referenced by multiple `PreviewInfo`'s. That's why need to be able to explicitly set `info.portInfo` here. 

@Nemikolh this is not exactly in line with your first review - let me know what you think!

Nemikolh:
> I think we need to either make it `Map<number, PortInfo[]>` or public portInfo, so that we can mark multiple PortInfo's ready at once.

I don't think I'm following that reasoning. A single instance of `PreviewInfo` uses a single port.

In your code snippet:

> ```ts
>    const previewInfos = previews.map((preview) => {
>      const info = new PreviewInfo(preview);
>      const portInfo = this._availablePreviews.get(info.port);
>
>      if (!portInfo) {
>        this._availablePreviews.set(info.port, info.portInfo);
>      } else {
>        info.portInfo = portInfo; // <-- All previews sharing same port should point to same portInfo
>      }
>
>      return info;
>    });
> ```

The logic should be changed to:

```ts
    const previewInfos = previews.map((preview) => {
      const port = PreviewInfo.parsePort(preview);
      let portInfo = this._availablePreviews.get(port);

      if (!portInfo) {
        portInfo = new PortInfo(port);
        this._availablePreviews.set(port, portInfo);
      }

      return new PreviewInfo(preview, portInfo);
    });
```

packages/runtime/src/webcontainer/preview-info.ts
```diff
@@ -15,27 +15,41 @@ export class PreviewInfo {
...
     } else if (Array.isArray(preview)) {
-      this.port = preview[0];
+      this.portInfo = new PortInfo(preview[0]);
       this.title = preview[1];
       this.pathname = preview[2];
     } else {
-      this.port = preview.port;
+      this.portInfo = new PortInfo(preview.port);
       this.title = preview.title;
       this.pathname = preview.pathname;
```
Nemikolh, line 46: I think we probably want to move out the parsing logic outside of the constructor. The constructor could accept everything as arguments and we have a static method that does the parsing and returns `{ port: number; pathname?: string; title?: string }` as an object.

AriPerkkio: Added `PreviewInfo.parse` that's used to parse the schema into preview's properties.

AriPerkkio:
> The logic should be changed to:

This looks good, I've applied these changes! 

This allows us to have multiple `PreviewInfo`'s referencing a single `PortInfo`, so that on webcontainer's `port` event we can mark multiple `PreviewInfo`'s ready at once. This is what https://github.com/stackblitz/tutorialkit/pull/233#issuecomment-2283265798 was essentially about. 👍

Nemikolh (approved):
This looks really good! 🤩 

Really good job on this! 🎉

PR #236, opened 2024-08-12
chore: remove spec files from runtime package

### LINKED ISSUE #158: `@tutorialkit/runtime` release includes test files
### Describe the bug

Published `@tutorialkit/runtime` package contains test files. Let's remove those.

_No response_

### Steps to reproduce

Check https://www.npmjs.com/package/@tutorialkit/runtime?activeTab=code

### Expected behavior

Test files should be excluded from published release. 

### Screenshots

_No response_

### Platform

- TutorialKit version: 0.0.2

### Additional context

_No response_

PhearZero:
# Overview

Looking at the package, everything seemed like it should not include the spec files. After some investigation I found some strange behavior with the components package being included in the build 🤔. 

# Summary

Package does not respect the `tsconfig.build.json` when `@tutorialkit/components-react` is included in the `--filter`.

# Fix

The quick and dirty fix is to just exclude the `@tutorial/components-react` from the filter and have a two step build. It may need further triage with pnpm/typescript. Happy to dig deeper! 

- resolves #158

### DIFF
--- a/packages/astro/tsconfig.json
+++ b/packages/astro/tsconfig.json
@@ -9,9 +9,9 @@
   },
   "include": ["src"],
   "references": [
-    { "path": "../runtime" },
-    { "path": "../types" },
-    { "path": "../components/react" },
+    { "path": "../runtime/tsconfig.build.json" },
+    { "path": "../types/tsconfig.build.json" },
+    { "path": "../components/react/tsconfig.build.json" },
     { "path": "../theme" }
   ]
 }
--- a/packages/components/react/tsconfig.build.json
+++ b/packages/components/react/tsconfig.build.json
@@ -7,5 +7,9 @@
   },
   "include": ["src"],
   "exclude": ["src/**/*.spec.ts"],
-  "references": [{ "path": "../../runtime" }, { "path": "../../theme" }, { "path": "../../types" }]
+  "references": [
+    { "path": "../../runtime/tsconfig.build.json" },
+    { "path": "../../theme" },
+    { "path": "../../types/tsconfig.build.json" }
+  ]
 }
--- a/packages/components/react/tsconfig.json
+++ b/packages/components/react/tsconfig.json
@@ -9,5 +9,9 @@
     "rootDir": "src"
   },
   "include": ["src"],
-  "references": [{ "path": "../../runtime" }, { "path": "../../theme" }, { "path": "../../types" }]
+  "references": [
+    { "path": "../../runtime/tsconfig.build.json" },
+    { "path": "../../theme" },
+    { "path": "../../types/tsconfig.build.json" }
+  ]
 }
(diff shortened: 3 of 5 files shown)


AriPerkkio:
Thanks for looking into this!

> Package does not respect the `tsconfig.build.json` when `@tutorialkit/components-react` is included in the `--filter`.

Oh wow, this is the case indeed. As soon as `@tutorialkit/components-react` is dropped from build script, test files are excluded. This seems to happen even in https://github.com/stackblitz/tutorialkit/pull/155 where the package was moved to `packages/react`, so it seems this is not caused by nested directories. 

It would be great to figure out why exactly this is happening. 🤔

PhearZero:
Anytime! Thank you all for the project ❤️. I was shocked as well!

>It would be great to figure out why exactly this is happening. 🤔

If someone doesn't get to it, I'll try to dig in this evening (it's going to be non-obvious). My first instinct is to dig into `pnpm` v8.15.6. I'll report back here with any findings 🕵️

Nemikolh:
Hey @PhearZero, thanks for investigating! This is indeed really odd.

I wonder if this is a bug with the fact that the package is marked as composite but project reference default to the `tsconfig.json`. I think if we modify the `{ "path": "..." }` references to point to the `tsconfig.build.json` it will work?

PhearZero:
Thanks for the hint @Nemikolh ❤️, I was able to update the `references` for the `runtime` to point to the `tsconfig.build.json` and that seems to have resolved the build!

Nemikolh (commented):
Looks good! Do you think you could update the reference everywhere? There's also the `tsconfig.json` in the same folder as the file you modified and the one in `astro`.

Those aren't used by the build script but they're used by VSCode and so I think it might also get those files to be transpiled? In any case, it's probably less error prone to always reference the correct config everywhere.

PhearZero:
@Nemikolh Makes total sense, I've updated the the references to point to the build configurations

AriPerkkio:
@PhearZero the CI is failing on lint step. Could you run `pnpm run lint --fix` and commit the changes.

Or check the `Maintainers are allowed to edit this pull request.` for this PR and I'll push the required changes.

AriPerkkio (approved):
Looks good to me! But Github won't allow me to merge this as the branch is not up-to-date with `main`. 

@PhearZero could you check this checkbox on from the right side of the page and I can proceed with this PR? 

image

PhearZero:
@AriPerkkio it looks like `edits by maintainer` is not supported in a [fork from another organization](https://github.com/orgs/community/discussions/5634)

I was able to use the **Update** branch button to merge in `main`

AriPerkkio:
Oh right, you forked TutorialKit repo to another Github organization instead of personal account. That makes sense.

Thanks for the fix!

Issue #235, opened 2024-08-09
Tutorial created in workspace fails


PhearZero:
### Describe the bug

It looks like vite is not resolving the dependencies correctly when it's in an npm workspace.

## Error

>Failed to resolve import "node_modules/astro/dist/transitions/swap-functions" from "node_modules/@tutorialkit/astro/dist/default/layouts/Layout.astro?v=c1a1fee0&astro&type=script&index=0&lang.ts". Does the file exist?

### Link to a StackBlitz project which shows the error

https://github.com/phearzero/withastro-astro-ru7fip

### Steps to reproduce

1. npm init
2. add `packages/*` to workspaces in package.json
2. mkdir packages
3. cd packages && npm create astro@latest -- --template starlight
4. cd ../packages && npm create tutorial
5. cd ../../ && npm install

If everything is properly deduplicated in the workspace, then tutorialkit is looking in the ./packages/tutorial/node_modules folder which only has the vite bundle.

### Expected behavior

Ability to use in a workspace that includes a `docs` module which consumes the workspace packages

### Platform

- TutorialKit version: 0.1.2
- OS: NA
- Browser: NA
- Version: NA

AriPerkkio:
The error seems to be coming from here: 

https://github.com/stackblitz/tutorialkit/blob/bd0ca1a5d1eb6a17ec2fcd6d776fe68458fd3cd6/packages/astro/src/default/layouts/Layout.astro#L41

PhearZero:
@Nemikolh no luck, looks like a similar problem on the assertion [here](https://github.com/stackblitz/tutorialkit/blob/745be37ef20ae97d6ded221fca24670742981879/packages/cli/src/commands/eject/index.ts#L152). I'll try to bypass it and test again

```bash
 ERROR  ./withastro-astro-ru7fip/packages/tutorial/node_modules/@tutorialkit/astro does not exists!
```

AriPerkkio:
We are waiting for a feature from Astro's side that allows us to remove the hacky private API solution from TutorialKit that is causing this bug. 

Meanwhile you can use following work-around in your `astro.config.ts`:

```ts
import { createRequire } from 'node:module';
import { resolve } from 'node:path';
import tutorialkit from '@tutorialkit/astro';
import { defineConfig } from 'astro/config';

const require = createRequire(import.meta.url);
const astroDist = resolve(require.resolve('astro/package.json'), '..');
[...]

PR #238, opened 2024-08-13
fix(astro): resolve Astro's private APIs properly

AriPerkkio:
- Fixes https://github.com/stackblitz/tutorialkit/issues/235

The usage of private APIs was added in https://github.com/stackblitz/tutorialkit/pull/4 (👋 @d3lm).


Nemikolh (commented):
This API was added after we talked to them 😄 

Having that import is intentional here. Once it's stabilized on the Astro side, we'll be able to get rid of the type error.

AriPerkkio:
> we'll be able to get rid of the type error

This is not just type error. In monorepo setups this import actually fails and TutorialKit fails to start. Here's example from https://github.com/AriPerkkio/tutorialkit/tree/test/ui-tests/test/ui which doesn't work without this fix.

d3lm:
Yes, I agree with @Nemikolh and I'd like to keep it as is. The whole point to get this in upstream is to avoid copying code from Astro so we can fully rely on upstream for this. I think we can give them the feedback that it works well for us so it can be exposed properly and we don't need to import from `node_modules` directly. But I am not a fan of copying it again cause that's what we wanted to avoid in the first place.

AriPerkkio:
Sure, that would be ideal case. It's also breaking change as we would need to set new minimum version of `astro` but we can do breaking change release at this point easily.

If you have any ideas how to fix #235 and https://github.com/stackblitz/tutorialkit/pull/241 without this change, feel free to help! Otherwise this will block those two, I think.

Nemikolh:
@AriPerkkio For #235, I think we can temporarily use a config option. We should definitely provide feedback that this has been working fine for us.

For #241, why is this an issue there? Doesn't the dependency also get installed by `pnpm`? (I only had a quick look so sorry if it's something obvious 🙈)

AriPerkkio:
Removed copied Astro's `swap-functions.js` and added Vite plugin that resolves it properly. I tried to add module resolving logic directly to `Layout.astro` but it just didn't work. It's best to do on Vite's side.

Tested with https://github.com/stackblitz/tutorialkit/issues/235 and https://github.com/stackblitz/tutorialkit/pull/241. Checked dev mode and build `dist` contents. 

Using private APIs like this relies on Astro's npm package's file structure and can break at any time. There's reason why `package.json`'s `exports` are required for imports. In my opinnion it would be best to simply copy-paste this from Astro if it's wanted to be used. And once they include it in their public API (`exports`), we can start requiring that version as peer dep.

d3lm:
> We should definitely provide feedback that this has been working fine for us.

@Nemikolh @AriPerkkio I can provide that feedback to them and ask them to expose it properly for us.

packages/astro/src/vite-plugins/astro-swap-functions.ts
```diff
@@ -0,0 +1,36 @@
...
+ * A plugin for importing Astro's private 'transitions/swap-functions.js' module.
+ * This is temporary solution and can break at any time when end-user's update their `astro` version.
+ */
+import { readFile } from 'node:fs/promises';
+import { createRequire } from 'node:module';
+import { resolve } from 'node:path';
+import type { VitePlugin } from '../types.js';
+
+const require = createRequire(import.meta.url);
+const astroDist = resolve(require.resolve('astro/package.json'), '..');
```
Nemikolh, line 11: Oh interesting, why do we `require.resolve` `astro/package.json` instead of just `astro`? 👀

packages/astro/src/vite-plugins/astro-swap-functions.ts
```diff
@@ -0,0 +1,36 @@
...
+      if (id === virtualModuleId) {
+        return resolvedId;
+      }
+
+      return undefined;
+    },
+
+    async load(id) {
+      if (id === resolvedId) {
+        return await readFile(swapFunctionEntry, 'utf8');
```
Nemikolh, line 30: Wouldn't it be better to return an import to that path? If the module import other files, those import will fail if we return its content like this.

packages/astro/src/default/layouts/Layout.astro
```diff
@@ -38,7 +38,8 @@ const baseURL = import.meta.env.BASE_URL;
       }
     </script>
     <script>
-      import * as builtInSwap from 'node_modules/astro/dist/transitions/swap-functions';
+      // @ts-ignore -- untyped
```
Nemikolh, line 41: We can add an entry for that module in `env-default.d.ts` to avoid the `//@ts-ignore`.

(we used to have a type error here though that we ignored so maybe this is fine)

Nemikolh (commented):
Looking good! 🤩

packages/astro/src/default/layouts/Layout.astro
d3lm, line 41:
```suggestion
      // @ts-ignore because it's untyped
```

packages/astro/src/vite-plugins/astro-swap-functions.ts
d3lm, line 1: Clever!

packages/astro/src/vite-plugins/astro-swap-functions.ts
AriPerkkio: I haven't heard of this problem before. How would that look in practice here? 

Is this error mentioned somewhere here https://rollupjs.org/plugin-development/#load?

packages/astro/src/vite-plugins/astro-swap-functions.ts
AriPerkkio: Resolving `astro` gives us the entrypoint of the package. Here we want to instead look exactly for `dist/transitions` directory.

```
> require.resolve("astro")
'/Users/x/tutorialkit/node_modules/.pnpm/astro@4.12.2_@types+node@20.14.11_typescript@5.5.3/node_modules/astro/dist/core/index.js'
> require.resolve("astro/package.json")
'/Users/x/tutorialkit/node_modules/.pnpm/astro@4.12.2_@types+node@20.14.11_typescript@5.5.3/node_modules/astro/package.json'
```

We could of course assume that their entrypoint always stays at `astro/dist/core/index.js`, and traverse the path up from there. 🤔

packages/astro/src/default/layouts/Layout.astro
AriPerkkio: I think it's better to keep `@ts-ignore`. By defining types in declaration file, we would need to somehow try to keep them up-to-date with Astro's implementation.

packages/astro/src/vite-plugins/astro-swap-functions.ts
Nemikolh: Ah thanks for the explanation!

packages/astro/src/vite-plugins/astro-swap-functions.ts
Nemikolh: Ah no worries! I'll clarify. 

Consider relative imports. There is no way they can work as the module path is lost when just returning the module content as a virtual module. If you look at what we did for the `theme.css` file, we import it (instead of including its content) precisely to support local imports.

packages/astro/src/default/layouts/Layout.astro
Nemikolh: Oh I was thinking of declaring the module as exporting `any` like we do for CSS modules.

packages/astro/src/vite-plugins/astro-swap-functions.ts
AriPerkkio: Oh good point, I didn't realize this! 💯 

I'll change this to `export * from '${swapFunctionEntry}';`. There's no `default` export so maybe we don't need to handle that one here.

packages/astro/src/default/layouts/Layout.astro
AriPerkkio: But that would require specifying each named export in the declaration, right? There's no way to say dynamically "all named exports from here are any".

Nemikolh: You don't have to. You can do like what we do for CSS modules:

```ts
declare module 'tutorialkit:astro-swap-functions';
```

AriPerkkio: Nice - happy to learn new tricks like this! 💯

d3lm:
Ok PR is on the way https://github.com/withastro/astro/pull/11708

Nemikolh:
@d3lm, @AriPerkkio do we want to wait for that PR to land on Astro's side then? Or are we all happy to move forward with this PR?

AriPerkkio:
Let's close this and wait for Astro's PR instead. I can add work-around into the #241 for now, and post same instructions in #235.

PR #239, opened 2024-08-13
docs: change homepage layout and add footer

sulco:


Nemikolh:
Looking great! I think on mobile we could keep the existing layout for the buttons though. IMO it was a bit better.

| before | after |
|--------|-------|
|    before |   after    |

What do you think?

PR #240, opened 2024-08-13
fix(astro): don't modify state during re-renders of `<WorkspacePanelWrapper />`

AriPerkkio:
Fixes React's state warning that happen when navigating between pages. `setLesson` that's done inside rendering block attempts to update state of the components that are just about to unmount.

```
Warning: Cannot update a component (`PreviewPanel`) while rendering a different component (`WorkspacePanelWrapper`). To locate the bad setState() call inside `WorkspacePanelWrapper`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
    at WorkspacePanelWrapper (http://localhost:4321/@fs/Users/x/tutorialkit/packages/astro/dist/default/components/WorkspacePanelWrapper.tsx:23:41)

printWarning                        @    react-dom.development.js:86
error                               @    react-dom.development.js:60
warnAboutRenderPhaseUpdatesInDEV    @    react-dom.development.js:27531
scheduleUpdateOnFiber               @    react-dom.development.js:25537
forceStoreRerender                  @    react-dom.development.js:16158
handleStoreChange                   @    react-dom.development.js:16134
notify                              @    index.js:55
set                                 @    index.js:72
setPreviews                         @    previews.js:77
setLesson                           @    index.js:96
WorkspacePanelWrapper               @    WorkspacePanelWrapper.tsx:15
```

Issue #209, opened 2024-08-04
Subfolders on windows do not work


noam-honig:
### Describe the bug

I'm using windows, I've added a file to a sub folder:
`src\content\tutorial\1-basics\1-introduction\1-welcome\_files\src\file-in-src.js`
Probably since I'm using windows it confuses it, instead of showing a folder and a file, it shows the full path,
and when I click it - I see an empty file.

![image](https://github.com/user-attachments/assets/227063ba-2ad8-4f54-91bc-878d72354a7f)

Checkout this repo:
https://github.com/noam-honig/play-with-tutorial-kit/

### Steps to reproduce

1. Use windows
2. Add a file called `src\content\tutorial\1-basics\1-introduction\1-welcome\_files\src\file-in-src.js`
3. Restart the dev server.
4. See that the file looks wrong and doesn't open

### Expected behavior

Expected to show a folder tree with files in it, and when the files are clicked - see that they open

### Screenshots

![image](https://github.com/user-attachments/assets/317b1677-58d1-47a5-96f2-3d7e0f22f2e7)

### Platform

- TutorialKit version: 0.1.2
- OS: Windows
- Browser: chrome
- Version: [e.g. 91.1]

AriPerkkio:
I can reproduce this on virtualized Windows 11 too. 

We should really add unit test for this and make sure it's run on Windows CI. I would start debugging from these files first:

https://github.com/stackblitz/tutorialkit/blob/cc50ad4d90d6ffb42f3a658de08f9617826b6c07/packages/runtime/src/lesson-files.ts#L12

https://github.com/stackblitz/tutorialkit/blob/cc50ad4d90d6ffb42f3a658de08f9617826b6c07/packages/astro/src/default/utils/content.ts#L81-L82

noam-honig:
FYI - in the meantime, I switch to `wsl` (linux on windows) and it seems ok there

AriPerkkio:
This is fixed in latest `0.1.4` release. Update the `@tutorialkit/*` packages in your `package.json` to latest version and Windows should work fine.

noam-honig:
Hi @AriPerkkio Tested - it works great

Issue #183, opened 2024-07-29
Preview window has height 0


noam-honig:
### Describe the bug

Tried to get started both with npm create tutorial and with `npx tutorialkit create` and in both cases I wasn't able to see the preview:
![image](https://github.com/user-attachments/assets/09147b80-9c2b-40c8-ba37-ec82e2124e45)

### Steps to reproduce

1. `npm create tutorial`
2. open browser
3. see missing preview.

### Expected behavior

I expected to see the preview - I also expected the default styling to be a bit better

### Screenshots

![image](https://github.com/user-attachments/assets/447642e0-9d36-475f-aea4-fd7a1979ea2c)

### Platform

- TutorialKit version: 0.1.0
- OS: Windows
- Browser: Chrome
- Version: latest

AriPerkkio:
This looks same as https://github.com/stackblitz/tutorialkit/issues/182. 

Could you try applying the fix from https://github.com/stackblitz/tutorialkit/pull/184/files on your `uno.config.ts`?

noam-honig:
Yeah - look like that did it

Issue #182, opened 2024-07-29
Initial setup (CSS) broken on Windows 


LukasSonnabend:
### Describe the bug

Hello I just setup a new Tutorialkit Project, unfortunately using Node 22.4.1/22.2.0 and running npx tutorialkit create the result is broken on Windows 11 (it seems some of the CSS is not loading). On my Mac with the same Node Version the CSS is working.

### Link to a StackBlitz project which shows the error

[Stackblitz Repro (not broken because not on windows?)](https://stackblitz.com/~/github.com/LukasSonnabend/tutorialkit-repro)

### Steps to reproduce

1. npx tutorialkit create
2. Yes to all
3. Visit link in Terminal

### Expected behavior

The initial State after running npx tutorialkit create should be working across OS.

### Screenshots

Found Behaviour
![grafik](https://github.com/user-attachments/assets/363f6106-1dad-4898-9a5a-a87a2fb638e8)

Expectation is the way the Demo looks: 
![grafik](https://github.com/user-attachments/assets/1a279083-a863-4a7a-8981-f212f63cc7dd)

### Platform

- TutorialKit version: 0.1.1
- OS: Windows 11
- Browser: Chrome, Edge, Firefox
- Version: 126.0.6478.128, 126.0.2592.113, 129.0b8 (64-bit)

AriPerkkio:
Could you try if following change in `uno.config.ts` fixes the issue?

```diff
-import { basename, dirname, join } from 'node:path';
+import { basename, dirname, join, sep } from 'node:path';

...

content: {
  inline: globSync([
-    `${convertPathToPattern(join(require.resolve('@tutorialkit/astro'), '..'))}/default/**/*.astro`,
-    `${convertPathToPattern(join(require.resolve('@tutorialkit/components-react'), '..'))}/**/*.js`,
+    `${convertPathToPattern(join(require.resolve('@tutorialkit/components-react'), '..')).replaceAll(sep, "/")}/**/*.js`,
+    `${convertPathToPattern(join(require.resolve('@tutorialkit/astro'), '..')).replaceAll(sep, "/")}/default/**/*.astro`,
  ]).map((filePath) => {
    return () => fs.readFile(filePath, { encoding: 'utf8' });
  }),
},
```

We are planning to provide better way to handle `uno.config.ts`, so that this much logic would not end up in end-users configurations.

PR #241, opened 2024-08-13
test: add tests for ui

AriPerkkio:
- Adds tests for UI using Playwright
- ~~Requires https://github.com/stackblitz/tutorialkit/pull/238~~
- Requires https://github.com/stackblitz/tutorialkit/pull/240

## Structure

Test cases are located in `test` directory.
Each test file has its own `chapter`, that contains `lesson`s for test cases:

For example Navigation tests:

```
├── src/content/tutorial
│   └── tests
│       └──── navigation
│           ├── page-one
│           ├── page-three
│           └── page-two
└── test
    └── navigation.test.ts
```

Or File Tree tests:

```
├── src/content/tutorial
│   └── tests
│       └── file-tree
│           └── lesson-and-solution
└── test
    └── file-tree.test.ts
```


packages/components/react/src/core/FileTree.tsx
```diff
@@ -162,14 +163,16 @@ interface ButtonProps {
...
 }
 
-function NodeButton({ depth, iconClasses, onClick, className, children }: ButtonProps) {
+function NodeButton({ depth, iconClasses, onClick, className, 'aria-pressed': ariaPressed, children }: ButtonProps) {
   return (
     <button
       className={`flex items-center gap-2 w-full pr-2 border-2 border-transparent text-faded ${className ?? ''}`}
       style={{ paddingLeft: `${12 + depth * NODE_PADDING_LEFT}px` }}
       onClick={() => onClick?.()}
+      aria-pressed={ariaPressed === true ? 'true' : undefined}
```
AriPerkkio, line 175: This isn't 100% correct either as these are not toggle buttons, but it's better than nothing. Proper way would be to throw these buttons away and create a proper tree: https://www.w3.org/WAI/ARIA/apg/patterns/treeview/

test/ui/test/file-tree.test.ts
```diff
@@ -0,0 +1,59 @@
...
+  await page.getByRole('button', { name: 'Solve', disabled: false }).click();
+  await expect(page.getByRole('button', { name: 'Reset' })).toBeVisible();
+
+  // solution files
+  for (const file of ['example.html', 'example.js']) {
+    await page.getByRole('button', { name: file }).click();
+    await expect(page.getByRole('button', { name: file, pressed: true })).toBeVisible();
+
+    // TODO: Figure out why this is flaky
+    await page.waitForTimeout(1_000);
```
AriPerkkio, line 33: Couldn't figure out what's the root cause here. Sometimes there was the content of previous file was shown. Here's picture that shows contents of `example.html` even though `example.js` is selected:

AriPerkkio:
@Nemikolh this is now ready for final review and is no longer blocked by #238.

As this contains couple of `fix` commits as well, it might be good to merge this without squashing so that the fixes will show up in release notes properly.

Nemikolh:
@AriPerkkio I think I can't push to main directly and I do not have access to other merge options. 

Note that this would be contrary to our conventions and they've been used to get more details about a commit by looking at the PR number associated with a commit. 

In the early days of TutorialKit it was possible but we then decided to always go through PRs.

Do you think you could submit a separate PR that contains all the fixes? Or multiple PRs if you prefer.

AriPerkkio:
Oh I see, other merge options are completely disabled. It would be good to have squashing enabled by default, but allow rebase when needed.

I'll split this PR and move to fixes into separate PRs. 👍

AriPerkkio:
Alright, good to go now. Other PRs are in `main` and this one is up-to-date.

.github/workflows/ci.yaml
```diff
@@ -32,6 +32,9 @@ jobs:
       - name: Lint
         run: pnpm lint
 
+      - name: Install Playwright Dependencies
+        run: pnpm exec playwright install chromium --with-deps
```
Nemikolh, line 36: I think it's a bit overkill to run playwright tests on every platform times node version. 

How about we create a separate e2e section that only runs on `ubuntu-latest` and only run the `e2e` tests?

test/cli/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "tutorialkit-integration",
+  "name": "@tutorialkit-cli/test-cli",
```
Nemikolh, line 2: I would keep the previous name and the `integrations` folder name TBH. 

I prefer reserving `test` (or `tests`) for unit tests.

Let's move the ui tests in a `e2e` folder at the root of the project.

package.json
```diff
@@ -30,6 +30,7 @@
     "eslint-plugin-astro": "^1.2.3",
     "husky": "^9.0.11",
     "is-ci": "^3.0.1",
+    "playwright": "^1.46.0",
```
Nemikolh, line 33: I think this could be moved in `e2e/package.json`? Or was it a specific reason it was required here?

package.json
```diff
@@ -16,7 +16,7 @@
     "demo": "pnpm run --filter=demo.tutorialkit.dev dev",
     "demo:build": "pnpm run build && pnpm run --filter=demo.tutorialkit.dev build",
     "lint": "eslint \"{packages,docs,extensions,integration}/**/*\"",
-    "test": "pnpm run --stream --filter=@tutorialkit/* test --run"
+    "test": "CI=true pnpm run --stream --filter='@tutorialkit/*' test"
```
Nemikolh, line 19: Oh looks like this `CI=true` is a leftover from local testing.

.npmrc
Nemikolh, line 1: What does this do?

test/ui/src/templates/default/index.mjs
```diff
@@ -0,0 +1,23 @@
+import http from 'node:http';
+import { readFileSync } from 'node:fs';
+
+const server = http.createServer((req, res) => {
+  if (req.url === '/' || req.url === '/index.html') {
+    res.writeHead(200, { 'Content-Type': 'text/html' });
+    res.end(readFileSync('./index.html', 'utf8'));
```
Nemikolh, line 7: We can only return the string `index` here as opposed to reading from a file.

test/ui/src/templates/default/index.mjs
```diff
@@ -0,0 +1,23 @@
...
+
+    return;
+  }
+
+  if (req.url === '/about.html') {
+    res.writeHead(200, { 'Content-Type': 'text/html' });
+    res.end(readFileSync('./about.html'));
```
Nemikolh, line 14: Same here, let's return `about`.

test/ui/src/templates/default/package.json
Nemikolh, line 1: Let's remove this file and do `node ./index.mjs` in the metadata.

Nemikolh (commented):
Overall looks good!

package.json
AriPerkkio: It was intentional here. Playwright won't run if `--run` is passed, and Vitest would start in watch-mode without it. But this change will be reverted once https://github.com/stackblitz/tutorialkit/pull/241#discussion_r1724721600 is done.

.npmrc
AriPerkkio: It's pnpm feature: https://pnpm.io/cli/run#shell-emulator

It's like `cross-env` but built-in directly in `pnpm`. Makes `ENV_VAR=value some_command` work on Windows too. 

I'll revert the `CI=true` from https://github.com/stackblitz/tutorialkit/pull/241#discussion_r1724736243, but maybe we should still leave this here? It provides consistent experience for Windows and other OS.

test/cli/package.json
AriPerkkio: I thought having just a single folder in the root would be more organized. We have similar structure here: https://github.com/vitest-dev/vitest

But sure, I'll revert these and create a `<root>/e2e` instead.

.github/workflows/ci.yaml
AriPerkkio: Yup, I'll change this. Though I wonder if we will miss Windows related bugs like below in future:
- https://github.com/stackblitz/tutorialkit/issues/209
- https://github.com/stackblitz/tutorialkit/issues/183
- https://github.com/stackblitz/tutorialkit/issues/182

Nemikolh: I think those issues could be captured differently. Like with an integration test that run a few simple sanity checks against the generated CSS. (like checking that a class is present for instance)

As long as the same JS / CSS and html are identical between platform, we can afford to only run UI tests on `ubuntu-latest`.

test/cli/package.json
Nemikolh: In WebContainer we have the following conventions:

 - `*.spec.ts` files in either `test` or in `src` folders are unit tests
 - `integrations` contains tests that verify the logic at the node level
 -  `e2e` contains tests that look at the entire webcontainer infra treating it as a black box (more or less)

I think what we're doing with our UI tests here is akin to e2e. We're testing the product in its entirety.

I find it nice to follow similar conventions here because:
 - That was the plan when TutorialKit was created.
 - Having similar conventions between projects makes context switching easier.

However, happy to discuss it with the rest of the team and decide with them if we want to do something different  for TutorialKit.

.npmrc
Nemikolh: Oh that's a really neat feature! :heart_eyes:  Let's keep it!!

.github/workflows/ci.yaml
AriPerkkio: > As long as the same JS / CSS and html are identical between platform

And the generated JSON's like in #209.

package.json
AriPerkkio: It was needed earlier when `pnpm exec playwright install chromium --with-deps` was run in the root. I've now added `working-directory: e2e` there and it seems to work. Removed this dep from root `package.json`.

.github/workflows/ci.yaml
Nemikolh: Oh yeah :100:

.github/workflows/ci.yaml
```diff
@@ -35,6 +35,23 @@ jobs:
...
+
+      - uses: ./.github/actions/setup-and-build
+
+      - name: Install Playwright Dependencies
+        run: pnpm exec playwright install chromium --with-deps
+        working-directory: e2e
+
+      - name: Test
+        run: pnpm test
+        working-directory: e2e
```
Nemikolh, line 53: I think we could remove those `working-directory` if we do:

```yaml
run: pnpm --filter=tutorialkit-e2e exec playwright install chromium --with-deps
```

and:

```yaml
run: pnpm test:e2e
````

File: e2e/README.md
Comment (Nemikolh) [inline e2e/README.md:1]: This file is really nice :sunglasses:

File: e2e/package.json [hunk 10]
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "tutorialkit-e2e",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "dev": "astro dev",
+    "start": "astro dev",
```
Nemikolh, line 7: Nit: we can probably remove those two scripts?

e2e/test/utils.ts
```diff
@@ -0,0 +1,33 @@
+import { readdirSync, readFileSync, existsSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+
+const TESTS_DIR = fileURLToPath(new URL('../src/content/tutorial/tests', import.meta.url));
```
Nemikolh, line 4: Oh that's a really nice way to get around platform path issues and be more esm idiomatic.

Nemikolh (approved):
Looks really good! :clap: 

Only have a minor comment on `ci.yaml` and let's get this in! :tada:

.github/workflows/ci.yaml
AriPerkkio: Sure I can apply this, but isn't `working-directory` a bit cleaner and simpler here?

e2e/package.json
AriPerkkio: We should leave at least one of these. When developing test cases it's really nice to get the server running when making changes to content and templates.

I think we could remove `start` here. I added it as I was running both dev/start from time to time - mixed conventions across projects.

.github/workflows/ci.yaml
Nemikolh: I find it slightly better because when you wonder what CI runs, you can just copy that command and execute it in your terminal without wondering in which folder you should be. It will work regardless of your current working directory (as long as it's a subfolder of the repo ofc).

So saying you always run `pnpm test:e2e` is shorter than saying "`pnpm test` but make sure to `cd ./e2e` ah wait, you need to `cd ../../e2e` because you're in `astro`"

PR #242, opened 2024-08-13
fix(extension): only match tutorialkit content and meta files

### LINKED ISSUE #237: TutorialKit extension's intellisense kicks off in non-tutorialkit projects
### Describe the bug

When TutorialKit extension is active, it seems to work on all files with frontmatter, even if it's not a tutorialkit project (e.g. an Astro blog), leading to false warnings and autocompletions:

image

_No response_

### Steps to reproduce

1. Create a new folder, with a markdown file inside
2. Open the folder in vscode with the tutorialkit extension active
3. Add some frontmatter into the markdown file
```
---
slug: 1
---
Hello
```
4. See tutorialkit-relevant warnings 
image

### Expected behavior

The warnings and completions should not appear in the non-tutorialkit projects.

### Screenshots

_No response_

### Platform

- TutorialKit extension version: 0.1.1
- Visual Studio Code version: 1.92.0

### Additional context

Looks like the `isTutorialKitWorkspace` (in [`extensions/vscode/src/utils/isTutorialKit.ts`](https://github.com/stackblitz/tutorialkit/blob/main/extensions/vscode/src/utils/isTutorialKit.ts)) function might be handy here:)

Nemikolh:
This PR makes sure that the autocompletion and validation for the frontmatter applies primarily to files that should be within a tutorialkit project. The heuristic used is very simple and not perfect so there could still be false positive. However, after this PR it will become very unlikely as one would need to create a markdown file named `meta.md` or `content.md(x)` under a `src/content/tutorial` path.

Note that files under `_files` and `_solution` are ignored as this could be a common situation where the extension would incorrectly perform validation and provide autocompletions.

Closes #237

### DIFF
--- a/extensions/vscode/src/language-server/index.ts
+++ b/extensions/vscode/src/language-server/index.ts
@@ -25,14 +25,7 @@ connection.onInitialize((params) => {
           {
             uri: 'https://tutorialkit.dev/reference/configuration',
             schema,
-            fileMatch: [
-              '**/*',
-
-              // TODO: these don't work
-              'src/content/*.md',
-              'src/content/**/*.md',
-              'src/content/**/*.mdx',
-            ],
+            fileMatch: ['volar-embedded-content://tutorialkit_frontmatter/**/*'],
             priority: SchemaPriority.Settings,
           },
         ],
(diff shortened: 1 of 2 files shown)


AriPerkkio (approved):
Tested locally in VSCode workspaces - works really well! 💯 

I think we should get this fix released soon as it improves DX a lot.

PR #253, opened 2024-08-16
refactor: split `<WorkspacePanel>` into 3 internal components

AriPerkkio:
- Refactors `<WorkspacePanel>` internally
- Working on `<WorkspacePanel>` is currently difficult as it contains so much logic. Everything is declared inside a single React component. Let's split this component into three smaller ones to improve readability and maintainability.

Tested manually but let's instead wait for https://github.com/stackblitz/tutorialkit/pull/241 just to be sure.

@Nemikolh any thoughts about this?


packages/components/react/src/Panels/WorkspacePanel.tsx
```diff
@@ -21,41 +17,80 @@ interface Props {
...
+        hasPreviews={hasPreviews}
+      />
+    </PanelGroup>
+  );
+}
 
-  const onFileSelect = useCallback((filePath: string | undefined) => {
-    tutorialStore.setSelectedFile(filePath);
-  }, []);
+function Editor({ theme, tutorialStore, hasEditor }: PanelProps) {
```
Nemikolh, line 87: Should we maybe name this `EditorSection`? My brain has a hard time when multiple components in the same codebase have the same name and are wrappers of the other :see_no_evil:

packages/components/react/src/Panels/WorkspacePanel.tsx
```diff
@@ -71,157 +106,147 @@ export function WorkspacePanel({ tutorialStore, theme }: Props) {
...
-    const unsubscribe = tutorialStore.lessonFullyLoaded.subscribe((loaded) => {
-      if (loaded && lesson.data.autoReload) {
-        /**
-         * @todo This causes some race with the preview where the iframe can show the "wrong" page.
-         * I think the reason is that when the ports are different then we render new frames which
-         * races against the reload which will internally reset the `src` attribute.
-         */
-        // previewRef.current?.reload();
-      }
-    });
```
Nemikolh, line 88: Btw, I think we should re-introduce this logic. It used to be broken because previews were loosing their state between lessons. Because now they are not re-parented anymore, I think this logic should work as expected.

Maybe before removing it, we should have a PR that add it back and do some checks that it works with a simple node http server that does not do livereload?

packages/components/react/src/Panels/WorkspacePanel.tsx
```diff
@@ -71,157 +106,147 @@ export function WorkspacePanel({ tutorialStore, theme }: Props) {
...
+    if (terminalPanelRef.current?.isCollapsed()) {
+      if (!terminalExpanded.current) {
+        terminalExpanded.current = true;
+        terminal.resize(DEFAULT_TERMINAL_SIZE);
+      } else {
+        terminal.expand();
+      }
     } else {
-      terminal.expand();
+      terminalPanelRef.current?.collapse();
```
Nemikolh, line 172: Oh the inlining is nice. Those `showTerminal` and `hideTerminal` functions were not passed as props anywhere. This is simpler. I like it!

packages/components/react/src/Panels/WorkspacePanel.tsx
```diff
@@ -71,157 +106,147 @@ export function WorkspacePanel({ tutorialStore, theme }: Props) {
...
+        onEditorChange={(update) => tutorialStore.setCurrentDocumentContent(update.content)}
+      />
+    </Panel>
+  );
+}
 
-      terminalExpanded.current = false;
-    }
-  }, [hideTerminalPanel]);
+function Preview({
```
Nemikolh, line 147: Maybe `PreviewsSection`? (with an `s` to `Preview` as there are zero to many)

Nemikolh (commented):
Why not! :star_struck: 

Given that part is probably less likely to change at this point, this is a nice change!

I would use different name than `Terminal`, `Editor` and `Preview` as it would make reading this code harder IMO as we already have components named like this.

packages/components/react/src/Panels/WorkspacePanel.tsx
```diff
@@ -21,41 +17,80 @@ interface Props {
...
-    tutorialStore.setSelectedFile(filePath);
-  }, []);
+function Editor({ theme, tutorialStore, hasEditor }: PanelProps) {
+  const [helpAction, setHelpAction] = useState<'solve' | 'reset'>('reset');
+  const selectedFile = useStore(tutorialStore.selectedFile);
+  const currentDocument = useStore(tutorialStore.currentDocument);
+  const lesson = tutorialStore.lesson!;
 
-  const onHelpClick = useCallback(() => {
+  function onHelpClick() {
```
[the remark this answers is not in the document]
AriPerkkio: It's unnecessary as `EditorPanel` and its `FileTab` are not using `React.memo`. They will re-render when ever their parent `WorkspacePanel` or `Editor` re-renders.

We should only use `useCallback` in cases where it's explicitly required. Adding them unnecessarily may lead to unintended complexity. What do you think?

packages/components/react/src/Panels/WorkspacePanel.tsx
AriPerkkio: Sure! For consistency I'll rename these 3 internal components to `EditorSection`, `PreviewSection` and `TerminalSection`.

packages/components/react/src/Panels/WorkspacePanel.tsx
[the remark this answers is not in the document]
Nemikolh: > We should only use `useCallback` in cases where it's explicitly required. Adding them unnecessarily may lead to unintended complexity. What do you think?

What unintended complexity are you referring to? Do you mean thinking about dependencies?

I tend to think that component should provide stable references to components they render regardless of whether those direct components use memo or not. Especially because those component might pass that props down their tree, and down the line there will likely be a memoized component that now is never memoized. Which leads to unexpected perf issues.

Additionally, I like that it makes explicit what a callback's dependencies are and sometimes you realised that they can be tweaked to have none (like when using a `setState` to access a value as opposed to accessing the state directly).

When `useCallback` isn't used then downstream components can't just benefit from adding `memo` and it becomes a lot more annoying to inspect every usage up the rendering tree.

If by default we use `useCallback` everywhere in component higher up in the tree then it's easier to use `memo` and improve perf when needed.

Curious to hear @d3lm thoughts on the topic.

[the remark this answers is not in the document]
AriPerkkio: Is there a performance issue that we are solving here, or just optimizing (which isn't actually working at the moment due to lack of `React.memo`) for future development?

> What unintended complexity are you referring to?

Unnecessary use of `useCallback` and `useMemo`. When they are used without `React.memo`, they don't provide any help. In React codebases you should not default to `useCallback` - use it only when it's required. I'm quite sure there was good explanation for this in https://overreacted.io/ when React `16.8` was released.

> downstream components can't just benefit from adding `memo` and it becomes a lot more annoying to inspect every usage up the rendering tree.

Yep, the `CodeMirrorEditor` here is doing exactly that. It's not expecting the callback props to be stable, so it's creating refs for them.

> If by default we use `useCallback` everywhere in component higher up in the tree then it's easier to use `memo` and improve perf when needed.

If we start doing this, should we also start to add other performance optimizations as well? No unstable default function arguments, `memo` and `useMemo` everywhere? Or just add https://react.dev/learn/react-compiler. These should be opt-in when needed.

packages/components/react/src/Panels/WorkspacePanel.tsx
AriPerkkio: I'll check this! So what kind of setup is required, and what used to be the erroneous state that was seen here? Add two lessons with Node servers on each and make sure correct content is shown when navigating between the lessons?

AriPerkkio: Yep, this works! Though there's a bug in iframe selector here. The `preview-panel` never contains any `iframe` elements as children - they are some how mirrored from the `#previews-container` there 🤔 

https://github.com/stackblitz/tutorialkit/blob/98ef05b828ff8f3ab45a49e62bf1a4b79e65acfc/packages/react/src/Panels/PreviewPanel.tsx#L52-L59

AriPerkkio: Continues at https://github.com/stackblitz/tutorialkit/pull/303

Nemikolh:
Weird that GitHub closes PR when a PR reference a comment from another PR :sweat_smile:

AriPerkkio:
Alright I think this is ready now! 🎉

packages/react/src/Panels/WorkspacePanel.tsx
```diff
@@ -72,156 +108,168 @@ export function WorkspacePanel({ tutorialStore, theme }: Props) {
...
-    if (!terminal) {
-      return;
-    }
-
-    if (!terminalExpanded.current) {
+    if (terminal && !terminalExpanded.current) {
       terminalExpanded.current = true;
       terminal.resize(DEFAULT_TERMINAL_SIZE);
-    } else {
+    } else if (terminal) {
```
Nemikolh, line 167: Nit: maybe instead of checking twice for `terminal`, we could have an:

```ts
if (!terminal) { 
  return; 
}
```

Nemikolh (approved):
Looks really good! I love the improvements, the code is a lot better! :pinched_fingers: 

Just a minor nit and let's get this merged :partying_face:

PR #254, opened 2024-08-18
fix(cli): `create-tutorial` not depending on `@tutorialkit/cli` package

ycs77:
Rename `tutorialkit` to `@tutorialkit/cli` missing in #153.

The PR #153 has been renamed the `@tutorialkit/cli` dependency in `package.json`, but no changed in `index.ts`, if now running locally will see the error: `Error: Cannot find module 'tutorialkit'`, so this PR is fixed this.

### DIFF
--- a/packages/create-tutorial/src/index.ts
+++ b/packages/create-tutorial/src/index.ts
@@ -4,6 +4,6 @@ import { spawnSync } from 'node:child_process';
 import { createRequire } from 'node:module';
 
 const require = createRequire(import.meta.url);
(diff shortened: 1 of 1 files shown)


AriPerkkio (approved):
Thanks, good catch! 💯 

The latest [release of `create-tutorial`](https://www.npmjs.com/package/create-tutorial?activeTab=code) still relies on `tutorialkit` package, so end-users won't be running into this issue. But it also means that they are stuck on previous versions. I'll publish new release tomorrow. 👍

PR #255, opened 2024-08-18
fix: missing preview i18n

ycs77:
### Problem

I'm currently using the tutorialkit version v0.1.5, if added some i18n texts:

*docs/demo/src/content/tutorial/meta.md*
```md
---
type: tutorial
mainCommand: ['npm run dev', 'Starting http server']
prepareCommands:
  - ['npm install', 'Installing dependencies']

# add the i18n texts
i18n:
  prepareEnvironmentTitleText: '啟動環境'
  previewTitleText: '預覽'
  toggleTerminalButtonText: '切換終端機'
---
```

The i18n will work on webcontainer starting:

螢幕擷取畫面 2024-08-18 153215

But not working on webcontainer started:

螢幕擷取畫面 2024-08-18 155256

### PR Added
* Add missing `Toggle Terminal` i18n text in started `<Preview>`
* Add missing `Preview` i18n text

### DIFF
--- a/docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx
+++ b/docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx
@@ -42,7 +42,7 @@ The `I18nText` type has the following shape:
 ```ts
 type I18nText = {
   /**
-   * Template for formatting a part. Variables: ${index} and ${title}.
+   * Template on how to format a part. Variables: ${index} and ${title}.
    *
    * @default 'Part ${index}: ${title}'
    */
@@ -53,7 +53,7 @@ type I18nText = {
    *
    * @default 'Edit this page'
    */
-  editPageText?: string
+  editPageText?: string,
 
   /**
    * Text of the WebContainer link.
@@ -70,12 +70,53 @@ type I18nText = {
   startWebContainerText?: string,
 
   /**
-   * Text shown on the call to action button to start webcontainer when boot was blocked
-   * due to memory restrictions.
+   * Text shown in the preview section when there are no steps to run and no preview to show.
    *
    * @default 'No preview to run nor steps to show'
    */
   noPreviewNorStepsText?: string,
+
+  /**
+   * Text shown on top of the file tree.
+   *
+   * @default 'Files'
+   */
+  filesTitleText?: string,
+
+  /**
+   * Text shown on top of the steps section.
+   *
+   * @default 'Preparing Environment'
+   */
+  prepareEnvironmentTitleText?: string,
+
+  /**
+   * Text shown on top of the preview section when `previews[_].title` is not configured.
+   *
+   * @default 'Preview'
+   */
+  defaultPreviewTitleText?: string,
+
+  /**
+   * Text for the toggle terminal button.
+   *
+   * @default 'Toggle Terminal'
+   */
+  toggleTerminalButtonText?: string,
+
+  /**
+   * Text for the solve button.
+   *
+   * @default 'Solve'
+   */
+  solveButtonText?: string,
+
+  /**
+   * Text for the reset button.
+   *
+   * @default 'Reset'
+   */
+  resetButtonText?: string,
 }
 
 ```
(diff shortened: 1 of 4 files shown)


AriPerkkio (commented):
The preview title is also configurable via `previews` option:

```yaml
---
type: tutorial
previews:
  - title: 預覽
    port: 5173
---
```

But it does make sense to be able to configure the default text via `i18n` as well.

ycs77:
> The preview title is also configurable via `previews` option:

Oh, I see.

---

When I want to do i18n, I intuitively look for replaceable values ​​in the i18n options.

packages/types/src/schemas/i18n.ts
```diff
@@ -56,6 +56,13 @@ export const i18nSchema = z.object({
    */
   prepareEnvironmentTitleText: z.string().optional().describe('Text shown on top of the steps section.'),
 
+  /**
+   * Text shown on top of the preview section.
+   *
+   * @default 'Preview'
+   */
+  previewTitleText: z.string().optional().describe('Text shown on top of the preview section.'),
```
Nemikolh, line 64: I would be happy to merge this PR if we rename that text to `defaultPreviewTitleText`. What do you think @AriPerkkio ?

packages/types/src/schemas/i18n.ts
```diff
@@ -56,6 +56,13 @@ export const i18nSchema = z.object({
    */
   prepareEnvironmentTitleText: z.string().optional().describe('Text shown on top of the steps section.'),
 
+  /**
+   * Text shown on top of the preview section.
```
Nemikolh, line 60:
```suggestion
   * Text shown on top of the preview section when `previews[_].title` is not configured.
```

Nemikolh (commented):
Really nice work! :star2: 

I think we can definitely land this. Thanks for the PR! :smiley:

packages/types/src/schemas/i18n.ts
AriPerkkio: I agree, `defaultPreviewTitleText` works better here. Let's use it.

AriPerkkio (changes requested):
We should also mention this in documentation:

https://github.com/stackblitz/tutorialkit/blob/c7bad203045b702afda3176cece645bee4d4f6e3/docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx?plain=1#L42-L81

packages/types/src/schemas/i18n.ts
ycs77: done

ycs77:
Wait a minute while I look at why the test failed.

AriPerkkio:
All tests are failing. We are pinning a dependency version of transitive dependency in test cases. Another dependency just updated and doesn't support that version. There's fix in https://github.com/stackblitz/tutorialkit/pull/261.

docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx
```diff
@@ -70,12 +70,53 @@ type I18nText = {
   startWebContainerText?: string,
 
   /**
-   * Text shown on the call to action button to start webcontainer when boot was blocked
-   * due to memory restrictions.
```
Nemikolh, line 74: Oh good catch! This comment is meant to be on top of `startWebContainerText?: string,`

Nemikolh (approved):
Looks good! I think the test failure is probably unrelated to this PR.

docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx
ycs77: So do we also need to update `i18n.ts`?

https://github.com/stackblitz/tutorialkit/blob/e45aa1523f82f71a406bdc63c98d7a0a7e277371/packages/types/src/schemas/i18n.ts#L25-L33

Nemikolh: Oh good catch! Yes we should! :smiley:

ycs77: done

Nemikolh:
@ycs77 You have a lint error! Once fixed, we can merge your PR :raised_hands:

PR #259, opened 2024-08-18
fix(astro/types): `webcontainer` should be a `Promise<WebContainer>`

Nemikolh:
This PR fixes the type of `webcontainer` in `@tutorialkit/astro/types` to be `Promise<WebContainer>` and not `WebContainer`.


AriPerkkio (approved):
Looks good to me! 👍

PR #266, opened 2024-08-20
fix(react): solve-button not working before lesson loads

AriPerkkio:
- Split from https://github.com/stackblitz/tutorialkit/pull/241

If Solve-button is clicked quickly enough, it does not reveal the solution files. It does change the button text to "Reset" which is not true in that case.

https://github.com/stackblitz/tutorialkit/blob/095ed570702d1b8de9370565b94103cd0740c408/packages/runtime/src/store/index.ts#L290-L295


Nemikolh (approved):
Good catch! :star_struck:

PR #267, opened 2024-08-20
fix(react): editor and preview missing accessible names

AriPerkkio:
- Split from https://github.com/stackblitz/tutorialkit/pull/241

Adds accessible names for editor and preview so that users with assistive technologies can identify the elements.

### DIFF
--- a/packages/components/react/src/Panels/PreviewPanel.tsx
+++ b/packages/components/react/src/Panels/PreviewPanel.tsx
@@ -160,6 +160,10 @@ function Preview({ preview, iframe, previewCount, first, last, toggleTerminal, i
     if (preview.url) {
       iframe.ref.src = preview.url;
     }
+
+    if (preview.title) {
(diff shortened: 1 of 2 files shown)


packages/components/react/src/Panels/PreviewPanel.tsx
```diff
@@ -160,6 +160,10 @@ function Preview({ preview, iframe, previewCount, first, last, toggleTerminal, i
     if (preview.url) {
       iframe.ref.src = preview.url;
     }
+
+    if (preview.title) {
+      iframe.ref.title = preview.title;
```
AriPerkkio, line 165: `title` in iframes is special case:

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#accessibility
- https://dequeuniversity.com/tips/provide-iframe-titles

Nemikolh (approved):
Nice :+1:

Issue #244, opened 2024-08-14
Vite CJS build warning caused by import of `unoCSSConfig` from `@tutorialkit/astro`


AriPerkkio:
### Describe the bug

When TutorialKit is started in development or build mode, Vite's CJS build entrypoint error is shown:

```
$ pnpm dev
...
The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.
```

### Steps to reproduce

```sh
$ pnpm create tutorial

$ pnpm dev
...
The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.

```

```diff
$ code uno.config.ts

- import { unoCSSConfig } from '@tutorialkit/astro';

export default defineConfig({
-  ...unoCSSConfig,
```

```sh
$ pnpm dev

# No warnings

```

### Expected behavior

No warning should be shown

### Platform

- TutorialKit version: 0.1.4

henrikvilhelmberglund:
Not sure why but seems to be caused by spreading `...unoCSSConfig` instead of using the variables from `import { rules, shortcuts, theme } from "@tutorialkit/theme";`

The UnoCSS config could also be made into a preset if that's easier:

`presetTutorialKit.ts`
```ts
import { definePreset } from "unocss"
import { rules, shortcuts, theme } from "@tutorialkit/theme";

export default definePreset(() => {
  return {
    name: 'presetTutorialKit',
    rules,
    shortcuts,
    theme
  }
})
```

`uno.config.ts` (removing `...unoCSSConfig`)
```ts
  presets: [
    presetUno({
      dark: {
        dark: '[data-theme="dark"]',
      },
    }),
    presetTutorialKit(),
    ```

AriPerkkio:
I think this is happening when `unocss` imports the `uno.config.ts`, and that file imports `'@tutorialkit/astro'` that has imports for Astro stuff. Not sure why `unocss` decides to use CJS here.

We are planning to move most, if not all (?) `uno.config.ts` inside `@tutorialkit/*` packages so that end-users wouldn't have to maintain that file. Ideally tutorials wouldn't even have to create `uno.config.ts` unless they explicitly wanted to extend that. There's some discussion on https://github.com/stackblitz/tutorialkit/issues/144.

PR #184, opened 2024-07-29
fix(cli): normalize windows paths for `fast-glob`

AriPerkkio:
- Fixes https://github.com/stackblitz/tutorialkit/issues/182
- Fixes https://github.com/stackblitz/tutorialkit/issues/183
- Ref. https://github.com/mrmlnc/fast-glob/issues/452

Issue #144, opened 2024-07-16
Roadmap to 1.0


Nemikolh:
Hey there! 

Thanks for showing interest in using TutorialKit! We are currently planning to do a 1.0 release and we're excited to share a list of the changes that we are intending to make before the release.

#### Planned Changes

* [x] :rotating_light:  Renaming `tutorialkit` to `@tutorialkit/cli`. So we can use that name for the vscode extension and also because the `npm create tutorial` command is the recommended way to create a new tutorial.
  * https://github.com/stackblitz/tutorialkit/pull/153
* [x] :rotating_light: Renaming `@tutorialkit/components-react` to `@tutorialkit/react` and move it from `components/react` to `react`. Simpler name and descriptive enough.
  * https://github.com/stackblitz/tutorialkit/pull/155
* [x] :rotating_light: Update `src/content/config.ts` to import the full schema from `@tutorialkit/types`.
  * https://github.com/stackblitz/tutorialkit/pull/156
* [x] :rotating_light: UnoCSS config:
  * [x] Use a deep merge strategy and move default presets and transformers in `@tutorialkit/theme`.
    * https://github.com/stackblitz/tutorialkit/pull/270
  * [x] Find a way to remove `content.inline`
    * https://github.com/stackblitz/tutorialkit/pull/270
* [x] :book: Document docs on how tow write your own component with `tutorialkit:store`
  * https://github.com/stackblitz/tutorialkit/pull/348
* [x] :book: Document  that `tutorialkit:core` provide low level access to webcontainer and is to be used only as a last resort
  * https://github.com/stackblitz/tutorialkit/pull/348
* [x] :book: Document that using `@tutorialkit/react` without TutorialKit is experimental
  * https://github.com/stackblitz/tutorialkit/pull/346 
* [x] Have `create-tutorial` depend on `"@tutorialkit/cli": "latest"` and release a fire and forget version.
  * https://github.com/stackblitz/tutorialkit/pull/146
* [x] Remove `.vscode/settings.json`.
  * https://github.com/stackblitz/tutorialkit/pull/173
* [x] Add TutorialKit extension to the list of recommended extension in the generated template.
  * https://github.com/stackblitz/tutorialkit/pull/172
* [x] Extension: 
  * [x] Frontmatter autocomplete:
    * #143 
  * [x] Support folder-based and config-based ordering
    * #223
  * [x] Part, chapter, lesson deletion
    * #223

#### Nice to have, some post 1.0.0 launch:

[...]

PR #270, opened 2024-08-21
feat!: simplify UnoCSS integration

AriPerkkio:
- Mentioned on https://github.com/stackblitz/tutorialkit/issues/144
- Fixes https://github.com/stackblitz/tutorialkit/issues/244
- Testing on https://github.com/AriPerkkio/tutorial-vite-plugin/pull/21

Moves all configuration from `uno.config.ts` behind `@tutorialkit/theme` and `@tutorialkit/astro` packages. Users are no-longer needed to have complex `uno.config.ts` files in their projects. In cases like https://github.com/stackblitz/tutorialkit/pull/184 we no longer need to [instruct users to modify their `uno.config.ts`](https://github.com/stackblitz/tutorialkit/releases/tag/0.1.1) - we can actually apply these in `@tutorialkit` packages and just release them. 

Adds a custom `defineConfig` to `@tutorialkit/theme`, that can be used to configure TutorialKit theme and add additional UnoCSS configurations:

```ts
// uno.config.ts
import { defineConfig } from "@tutorialkit/theme";

export default defineConfig({
 // add your custom UnoCSS config here. See https://unocss.dev/guide/config-file
});
```

UnoCSS VS Code extension still works as expected:

image

## Migration guide for projects:

In your `package.json`:

```diff
  "dependencies": {
    "@tutorialkit/components-react": "0.1.5",
    "react": "^18.3.1",
    "react-dom": "^18.3.1"
  },
  "devDependencies": {
    "@astrojs/check": "^0.7.0",
    "@astrojs/react": "^3.6.0",
-   "@iconify-json/ph": "^1.1.13",
-   "@iconify-json/svg-spinners": "^1.1.2",
    "@tutorialkit/astro": "0.1.5",
+   "@tutorialkit/theme": "0.1.5",
    "@tutorialkit/types": "0.1.5",
    "@types/node": "^20.14.6",
    "@types/react": "^18.3.3",
-   "@unocss/reset": "^0.59.4",
    "astro": "^4.12.0",
-   "fast-glob": "^3.3.2",
    "prettier-plugin-astro": "^0.14.1",
    "typescript": "^5.4.5",
-   "unocss": "^0.59.4"
  }
```

In your `uno.config.ts`:

```diff
+ import { defineConfig } from '@tutorialkit/theme';
- import { unoCSSConfig } from '@tutorialkit/astro';
- import { globSync, convertPathToPattern } from 'fast-glob';
- import fs from 'node:fs/promises';
- import { basename, dirname, join } from 'node:path';
- import { defineConfig, presetIcons, presetUno, transformerDirectives } from 'unocss';

+ export default defineConfig({
+  // add your custom UnoCSS config here. See https://unocss.dev/guide/config-file
+ });
- const iconPaths = globSync('./icons/languages/*.svg');
-
- const customIconCollection = iconPaths.reduce(
-   (acc, iconPath) => {
-     const collectionName = basename(dirname(iconPath));
-     const [iconName] = basename(iconPath).split('.');
-
-     acc[collectionName] ??= {};
-     acc[collectionName][iconName] = async () => fs.readFile(iconPath, 'utf8');
-
-     return acc;
-   },
-   {} as Record<string, Record<string, () => Promise<string>>>,
- );

- export default defineConfig({
-   ...unoCSSConfig,
-   content: {
-     inline: globSync([
-       `${convertPathToPattern(join(require.resolve('@tutorialkit/components-react'), '..')).replace('\\@', '/@')}/**/*.js`,
-       `${convertPathToPattern(join(require.resolve('@tutorialkit/astro'), '..')).replace('\\@', '/@')}/default/**/*.astro`,
-     ]).map((filePath) => {
-       return () => fs.readFile(filePath, { encoding: 'utf8' });
-     }),
-   },
-   transformers: [transformerDirectives()],
-   presets: [
-     presetUno({
-       dark: {
-         dark: '[data-theme="dark"]',
-       },
-     }),
-     presetIcons({
-       collections: {
-         ...customIconCollection,
-       },
-     }),
-   ],
- });
```

## BREAKING CHANGES:
- `@tutorialkit/astro` package no longer exports `unoCSSConfig`. Use `defineConfig` from `@tutorialkit/theme` instead.
- `@tutorialkit/theme` package no longer exports `rules`, `shortcuts` and `theme`.
- `@tutorialkit/theme` package exports `transitionTheme` from new entrypoint: `@tutorialkit/theme/transition-theme`.


AriPerkkio:
@Nemikolh what do you think about adding https://pkg.pr.new to TutorialKit? Verifying changes like this would be really great if I could use real published packages. Relying on linked (`file://` or `link://` protocol on `package.json`) doesn't provide good confidence. Package managers and Node resolving is weird.

Nemikolh:
Oh that would be really cool! I like a lot that idea :smiley:

packages/astro/src/integrations.ts
```diff
@@ -47,7 +49,14 @@ export function extraIntegrations() {
     mdx(),
     UnoCSS({
       configDeps: ['./theme.ts'],
-      injectReset: true,
+      injectReset: createRequire(root).resolve('@unocss/reset/tailwind.css'),
```
Nemikolh, line 52: Oh I'm curious: why is the root needed? Is it because `@unocss/reset` is not a dependency of `@tutorialkit/astro`? 

What about we add it as a dependency?

Or we could also keep the `true` value.

Nemikolh (approved):
Wooow, impressive work! This looks insanely good. :heart_eyes: 

Good job :smiley:

packages/theme/src/transition-theme.ts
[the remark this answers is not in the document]
AriPerkkio: I moved them to the entrypoint so that we can have `import { defineConfig } from "@tutorialkit/theme";` instead of `import { defineConfig } from "@tutorialkit/theme/config";`.

packages/astro/src/integrations.ts
AriPerkkio: > Is it because `@unocss/reset` is not a dependency of `@tutorialkit/astro`?

It is a dependency of `@tutorialkit/astro`. 

When `injectReset: true` is used I'm seeing following error:

image

packages/theme/src/transition-theme.ts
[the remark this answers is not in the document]
Nemikolh: Oh yeah that will probably be the way this is used the most. Yeah let's keep it this way :+1:

packages/astro/src/integrations.ts
Nemikolh: Oh that is strange, could that be a bug in the unocss astro integration?

packages/astro/src/integrations.ts
```diff
@@ -47,7 +49,14 @@ export function extraIntegrations() {
...
+      content: {
+        inline: getInlineContentForPackage({
+          name: '@tutorialkit/astro',
+          pattern: '/dist/default/**/*.astro',
+          root,
+        }),
+      },
```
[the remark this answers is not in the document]
AriPerkkio: Yep it seems so! Though I didn't find any documentation for this - I just expected it to work similar as other tools. 

When this part is commented out and locally linked project is **built and run with `pnpm preview`**, the styles break. When this is added back, it works. Note that when using `pnpm dev` this has no effect.

packages/theme/src/transition-theme.ts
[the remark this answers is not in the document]
AriPerkkio: Added comment here describing why this is a separate module.

packages/astro/src/integrations.ts
AriPerkkio: This is also mentioned on UnoCSS:

- https://github.com/unocss/unocss/issues/2342#issuecomment-1463455572

Now that we don't require end-users (or template) to add `@unocss/reset` as their dependency, its likely causing a hoisting issue here, like mentioned on the issue. Using `require.resolve` solves this.

Nemikolh: Gotcha, thanks for the clarification!

Nemikolh:
@AriPerkkio Not that I can think of! We can always fix them later once we encounter them. Let's move forward with this PR

PR #271, opened 2024-08-21
ci: setup pkg.pr.new

AriPerkkio:
- Adds https://pkg.pr.new to TutorialKit repo
- All commits pushed to `main` trigger an automatic release
- Maintainers can trigger a release by commenting `/pkg-pr-new` in a PR

This will make testing PRs like https://github.com/stackblitz/tutorialkit/pull/270 a lot easier. In that PR we need to verify that packages are resolved correctly in real environments. Testing such changes by linking packages with `file://` or `link://` protocols in `package.json` doesn't provide good confidence - using real published packages is better.

I've used the workflow from Vite as reference: https://github.com/vitejs/vite/blob/main/.github/workflows/publish-commit.yml

### DIFF
--- a/.github/workflows/publish-commit.yaml
+++ b/.github/workflows/publish-commit.yaml
@@ -1,0 +1,71 @@
+# Publishes main-branch's commits to pkg.pr.new
+# PRs can be published by commenting `/pkg-pr-new` in the PR
+
+name: Publish Any Commit
+
+on:
+  push:
+    branches:
+      - main
+  issue_comment:
+    types: [created]
+
+jobs:
+  build:
+    if: github.repository == 'stackblitz/tutorialkit' && (github.event_name == 'push' || github.event.issue.pull_request && startsWith(github.event.comment.body, '/pkg-pr-new'))
+    runs-on: ubuntu-latest
+
+    steps:
+      - if: github.event.issue.pull_request
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const user = context.payload.sender.login
+            console.log(`Validate user: ${user}`)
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-commit.yaml
```diff
@@ -0,0 +1,71 @@
...
+      - uses: ./.github/actions/setup-and-build
+
+      - name: Publish to pkg.pr.new
+        run: >
+          pnpm dlx pkg-pr-new@0.0 publish --compact --pnpm
+          ./packages/astro
+          ./packages/components/react
+          ./packages/runtime
+          ./packages/theme
+          ./packages/types
```
AriPerkkio, line 71: @Aslemammad I would rather use `--filter='@tutorialkit/*'` here - is it possible? There are some packages in `./packages/` that we don't want to publish - only the ones that have `@tutorialkit/*` in the name.

Similar to:

https://github.com/stackblitz/tutorialkit/blob/74e052701934fbbff9d8bcd06dafc88c58b83667/.github/workflows/publish-release.yaml#L29-L34

Aslemammad: i mean, we support globs, you can see the cli/index.ts file in pkg.pr.new! So if that's valid glob, it'd work.

AriPerkkio: Is it glob for filenames and paths, or for `name`'s in `package.json`?

AriPerkkio:
Verified the same script in separate repo: https://github.com/AriPerkkio/eslint-remote-tester/pull/549

PR #273, opened 2024-08-24
chore: rename import `@tutorialkit/test-utils` to `test-utils`

Nemikolh:
This PR reverts the renaming of the import back to how it was: `test-utils` (was renamed to `@tutorialkit/test-utils` in #141).

This change aligns better with conventions we have in other repositories and also makes it easier to find what the import correspond to given the name of the package is `test-utils` and not `@tutorialkit/test-utils`. Note that we could instead change that package name to be `@tutorialkit/test-utils` but this is not a desirable change we want given it's not meant to be published.

### DIFF
--- a/packages/runtime/src/tutorial-runner.spec.ts
+++ b/packages/runtime/src/tutorial-runner.spec.ts
@@ -1,8 +1,8 @@
 // must be imported first
-import { resetProcessFactory, setProcessFactory } from '@tutorialkit/test-utils';
+import { resetProcessFactory, setProcessFactory } from 'test-utils';
 
 import { WebContainer } from '@webcontainer/api';
-import type { MockedWebContainer } from '@tutorialkit/test-utils';
+import type { MockedWebContainer } from 'test-utils';
 import { beforeEach, describe, expect, test, vi } from 'vitest';
 import { TerminalStore } from './store/terminal.js';
 import { TutorialRunner } from './tutorial-runner.js';
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -11,7 +11,7 @@
     "noImplicitReturns": true,
     "strict": true,
     "paths": {
-      "@tutorialkit/test-utils": ["./packages/test-utils/src/index.ts"]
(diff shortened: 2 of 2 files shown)


AriPerkkio:
Internal packages should always use scoped package names. This is a security feature. 

But if that convention hasn't been followed else where, let's not follow it here either.

AriPerkkio (approved):

Nemikolh:
@AriPerkkio How is it a security feature?

AriPerkkio:
> ```diff
> - import { resetProcessFactory, setProcessFactory } from '@tutorialkit/test-utils';
> + import { resetProcessFactory, setProcessFactory } from 'test-utils';
> ```

This could import `test-utils` npm package, not our own internal one. In larger organizations it's quite common that all internal packages are required to be scoped ones, so that it's easy to detect which ones can be trusted. This post might describe the idea better: https://github.blog/security/supply-chain-security/avoiding-npm-substitution-attacks/#tldr

But well, if this kind of convention isn't followed in other packages, it's fine to not follow it here either.

AriPerkkio:
I recommend to read this one as well if you haven't seen it earlier: https://medium.com/@alex.birsan/dependency-confusion-4a5d60fec610
It's a bit longer post but really good. 💯

Nemikolh:
Oh thanks for the pointers! Maybe we could use an import path that cannot be published then like starting with `~` or something. Let's discuss it with everyone else on Monday then.

I don't like using the `@tutorialkit` prefix because it's confusing as where this is coming from and I don't want the package to be accidentally published.

Nemikolh:
@AriPerkkio I finished reading both, thanks again for sharing! 🤩

The behaviour of `--extra-index-url` in `python` is wild 😱

AriPerkkio:
>  like starting with `~` or something. 

That would work. I've seen `~`, `@/` and `#` prefixes used a lot for this. The `#` seems to be related to [sub path imports](https://nodejs.org/api/packages.html#subpath-imports) but I haven't used that before. 🤔

PR #277, opened 2024-08-26
docs: tutorialkit.dev link in readme

o-az:
https://www.tutorialkit.dev (with www) doesn't work


AriPerkkio (approved):
Good catch, thanks! 💯

Issue #287, opened 2024-08-28
Component docs contain references to excluded files


AriPerkkio:
### Describe the bug

In terminal example at https://tutorialkit.dev/reference/react-components/ there is import for `useTheme.ts` file that is not visible in the page, or in the linked Stackblitz project. There's no way for users to know what it contains.

```ts
import { useTheme } from './hooks/useTheme';
```

### Steps to reproduce

https://tutorialkit.dev/reference/react-components/

### Expected behavior

Documentation should contain all code that is required to reproduce the example. Ideally users should not even open the linked Stackblitz project to see what the referenced files contain. 

### Platform

- TutorialKit version: 0.2.0

Nemikolh:
We could introduce that file as a code snippet somewhere at the start of the page. In particular, it's not been shown because the actual implementation used in the docs is specific to our docs and not relevant for end users.

PR #289, opened 2024-08-28
docs: show sources for referenced files

AriPerkkio:
- Fixes https://github.com/stackblitz/tutorialkit/issues/287

### DIFF
--- a/docs/tutorialkit.dev/src/content/docs/reference/react-components.mdx
+++ b/docs/tutorialkit.dev/src/content/docs/reference/react-components.mdx
@@ -4,6 +4,7 @@ description: Use TutorialKit's React components for fun and profit.
 ---
 import PackageManagerTabs from '@components/Tabs/PackageManagerTabs.astro'
 import { Code } from '@astrojs/starlight/components';
+import { Tabs, TabItem } from '@astrojs/starlight/components';
 import themeCSS from '@tutorialkit/astro/default-theme.css?raw';
 
 import Example from '@components/react-examples/Example.astro';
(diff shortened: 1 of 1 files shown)


docs/tutorialkit.dev/src/content/docs/reference/react-components.mdx
```diff
@@ -17,6 +18,8 @@ import sourceExampleSimpleEditor from '@components/react-examples/ExampleSimpleE
 import sourceExampleFileTree from '@components/react-examples/ExampleFileTree?raw';
 import sourceExampleCodeMirrorEditor from '@components/react-examples/ExampleCodeMirrorEditor?raw';
 import sourceExampleTerminal from '@components/react-examples/ExampleTerminal?raw';
+import sourceUseTheme from "@components/react-examples/hooks/useTheme?raw";
```
Nemikolh, line 21: I would use a different version from that file. That file is very specific to our docs.

Here's what we could show for `hooks/useTheme.ts`:

```ts
export function useTheme() {
  return 'light';
}
```

AriPerkkio: We could also remove the hook usage completely and just use static `"light"` string as `theme` in the source code. What do you think?

Nemikolh: Hmm but then the docs wouldn't look as good when using viewing them in dark theme.

What drawback do you see in showing a different version for `useTheme`? I don't expect that hook to be changed very often.

docs/tutorialkit.dev/src/content/docs/reference/react-components.mdx
```diff
@@ -74,6 +77,17 @@ See a full example on [StackBlitz][stackblitz-example].
 
 ## Components
 
+These examples have references to following files:
```
Nemikolh, line 80:
```suggestion
These examples have references to the following files:
```

docs/tutorialkit.dev/src/content/docs/reference/react-components.mdx
Nemikolh: Ah no worries! :raised_hands:

Nemikolh (approved):
Nice! :star2:

Issue #290, opened 2024-08-29
ReferenceError: `__WC_CONFIG__` is not defined


AriPerkkio:
### Describe the bug

Freshly created tutorials without any modifications are running into `__WC_CONFIG__ is not defined` errors. This is coming up randomly and looks like possible race condition. When page is refreshed with cache clearing the error disappears, but normal refreshes run into this error immediatelly. 

This only affects dev mode. 

### Steps to reproduce

1. `pnpm create tutorial`
2. Start in dev mode
3. Open page in browser
4. Keep refreshing page and check JS console

### Expected behavior

Error should not happen. 

### Screenshots

![image](https://github.com/user-attachments/assets/1c83ad02-aeb5-4cf6-9e1b-a30c57094785)

### Platform

- TutorialKit version: 0.2.0

Nemikolh:
Could this be a bug with Astro? I wonder if we also see this issue with an older version of Astro.

I'm wondering because this is literally the first thing we set in the vite config and I don't think this has ever really changed:

https://github.com/stackblitz/tutorialkit/blob/3e7e6cb59f7f6ffd4551f483d71f36f51fcf0ef9/packages/astro/src/index.ts#L77-L94

(also if that helps bisecting the combo that breaks, I've never encountered that issue when I was doing demos back in June)

AriPerkkio:
This issue isn't present on `0.1.2` but appears first in `0.1.3`.

https://github.com/stackblitz/tutorialkit/compare/0.1.2...0.1.3

AriPerkkio:
This is likely introduced in https://github.com/stackblitz/tutorialkit/pull/219.

When import of `'./OpenInStackblitzLink.astro'` is removed here, the issue disappears:

https://github.com/stackblitz/tutorialkit/blob/3e7e6cb59f7f6ffd4551f483d71f36f51fcf0ef9/packages/astro/src/default/components/TopBarWrapper.astro#L6-L8

There's two depedency chains for the `setup.ts`:

```
TopBarWrapper.astro > setup
TopBarWrapper.astro > OpenInStackblitzLink.astro > webcontainer > setup
```

The first one works fine. But the second on comes from `<script>` tag. Maybe Astro doesn't apply Vite's `define` transforms there? I'll keep looking... 👀

AriPerkkio:
> Maybe Astro doesn't apply Vite's define transforms there? I'll keep looking...

Something weird going on there. Tested this in minimal Astro reproduction:

[...]

PR #293, opened 2024-08-30
fix(astro): work-around for dev-mode's `ReferenceError: __WC_CONFIG__ is not defined` errors

AriPerkkio:
- Fixes https://github.com/stackblitz/tutorialkit/issues/290
- Ref. https://github.com/withastro/astro/issues/11874

As work-around for the upstream issue, let's convert this component into React component instead of Astro one. It's then loaded on client side after Vite's `define` is applied, it seems. 

Note that the original bug is not visible when developing TutorialKit in the monorepo. It only appears in projects that are using the actual packages. Tested locally by linking packages with `pnpm`.


Nemikolh (approved):
Nice workaround! :smiley: Hopefully this gets fixed upstream.

PR #301, opened 2024-09-02
fix: align `Powered by WebContainers` to the bottom

AriPerkkio:
The `Powered by WebContainers` looks a bit weird as it's not aligned to the bottom of the viewport. @sulco & @Nemikolh - any thoughts?

Before:

After:


Nemikolh (approved):
Good catch! Yep much better 🙌

PR #305, opened 2024-09-02
feat(react): add button to reload a preview

### LINKED ISSUE #248: A small refresh button in the preview tab

There are cases where I want my student to be able to manually hit refresh on the preview window to reload it - this is espcially useful with the new `pathname` option for previews.

A small refresh button :) it can be optional and determined for each preview - in the previews config

I created my own "virutal" tabs and added my own refresh buttons, but it's not as cool

### Additional context

_No response_

Nemikolh:
This PR replaces the globe icon with a reload button, to reload a preview. It uses the new `reloadPreview` function from `@webcontainer/api/utils`.

Closes #248

<p align="center">

</p>

Note that this is best effort. Meaning that if the learner navigates within the preview to a location that does not render HTML (like an image or JSON), pressing the reload icon won't work as it will reset the iframe to its initial location. This can only be fixed at the WebContainer level.


e2e/src/templates/default/index.mjs
```diff
@@ -1,9 +1,10 @@
 import http from 'node:http';
+import fs from 'node:fs';
 
 const server = http.createServer((req, res) => {
   if (req.url === '/' || req.url === '/index.html') {
     res.writeHead(200, { 'Content-Type': 'text/html' });
-    res.end('Index page');
+    res.end(fs.readFileSync('./index.html', 'utf-8'));
```
AriPerkkio, line 7: On `main` there's now a `template: file-server` that does this. Originally this default one was doing it also but it was reverted to minimal no-fs-operations in https://github.com/stackblitz/tutorialkit/pull/241#discussion_r1724743673.

e2e/test/preview.test.ts
```diff
@@ -13,6 +13,22 @@ test('user can see single preview tab', async ({ page }) => {
   await expect(preview.getByText('Index page')).toBeVisible();
 });
 
+test('user can reload the page', async ({ page }) => {
```
AriPerkkio, line 16: `page` doesn't sound right here. Maybe something like:

```suggestion
test('user can reload a preview tab', async ({ page }) => {
```

e2e/test/preview.test.ts
```diff
@@ -13,6 +13,22 @@ test('user can see single preview tab', async ({ page }) => {
...
+
+  const preview = page.frameLocator('[title="Node Server"]');
+
+  await expect(preview.getByText('Index page')).toBeVisible();
+
+  await page.getByTestId('write-to-file').click();
+
+  await expect(preview.getByText('Index page')).toBeVisible();
+
+  await page.getByTitle('Reload Preview').click();
```
AriPerkkio, line 27: `byTitle` is good, but we can improve this. Let's also verify the element's role:

```suggestion
  await page.getByRole('button', { name: 'Reload Preview' }).click();
```

Now the fact that the name comes from `title` is just an implementation detail, and we can verify that proper element/role is used for this clickable element.

packages/types/src/schemas/i18n.ts
```diff
@@ -66,6 +66,16 @@ export const i18nSchema = z.object({
    */
   defaultPreviewTitleText: z.string().optional().describe('Text shown on top of the preview section.'),
 
+  /**
+   * Title or tooltip shown when hovering the reload icon for a preview.
```
AriPerkkio, line 70: Consider: I would like to avoid using the word `tooltip` here, or mention anything about hovering. Maybe something like:

```suggestion
   * Title attribute for the preview reload button.
```

e2e/src/components/ButtonWriteToFile.tsx
```diff
@@ -0,0 +1,27 @@
+import tutorialStore from 'tutorialkit:store';
+
+interface Props {
+  filePath: string;
+  newContent: string;
+  testId?: string;
+}
+
+export function ButtonWriteToFile({ filePath, newContent, testId = 'write-to-file' }: Props) {
```
AriPerkkio, line 9: Nice! I can imagine this utility component being useful in other future tests as well. It's good to have it accepting props like this. 🎉

AriPerkkio (commented):
Overall looks good! Great to have test case for this as well!

The background color inside the icon seems a bit weird to me. But after I noticed it, I saw that we already have similar background in Reset-button.

image

e2e/src/templates/default/index.mjs
Nemikolh: Ah nice! I'll use that one :+1:

henrikvilhelmberglund:
The duotone icon looks strange to me too, https://icones.js.org/collection/ph?s=arrow&icon=ph:arrow-clockwise could be used instead, along with a `dark:text-slate-100` or something if necessary for the color.

Nemikolh:
@henrikvilhelmberglund I'll change it to the `arrow-clockwise` then. Thanks for voicing your opinion :smiley:

AriPerkkio (approved):
Documentation for the new localization key should be added in `docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx`. Once that's done, this is good to go!

We should create separate follow-up issue for updating the background of the old icon-buttons too.

Nemikolh:
> We should create separate follow-up issue for updating the background of the old icon-buttons too.

@AriPerkkio Oh do you mean updating all the `ph-duo` icons for their non-duotone version?

AriPerkkio:
> @AriPerkkio Oh do you mean updating all the `ph-duo` icons for their non-duotone version?

Yep, at least the reset button looks similarly weird now that I've seen how clear the refresh button looks after the change. 😄 

Or maybe this in intentional?

image

Nemikolh:
@AriPerkkio It was, but we can also change them I think. It was not my choice :man_shrugging: 

Tbqh, I've always felt that, at the very least, they should be customizable. I'll open a ticket for that and another one to change them to their non-duotone versions.

PR #307, opened 2024-09-03
docs: change the clone url

noam-honig:
the git@github.com didn't work on my machine

Also - when trying to set this up, it required the specific pnpm@8.15.6 for some reason - so I've added that to the document

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,7 @@ Hi! We are really excited that you are interested in contributing to TutorialKit
 
 ## Repo Setup
 
-The TutorialKit repo is a monorepo using pnpm workspaces. The package manager used to install and link dependencies must be [pnpm](https://pnpm.io/). pnpm version 8.15.6 is required (npm i -g pnpm@8.15.6)
+The TutorialKit repo is a monorepo using pnpm workspaces. The package manager used to install and link dependencies must be [pnpm](https://pnpm.io/). Package manager versioning is handled by [`packageManager`](https://nodejs.org/api/packages.html#packagemanager) field that's supported by [Node's Corepack](https://nodejs.org/api/corepack.html) and other tools.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -4,14 +4,14 @@ Hi! We are really excited that you are interested in contributing to TutorialKit
 
 ## Repo Setup
 
-The TutorialKit repo is a monorepo using pnpm workspaces. The package manager used to install and link dependencies must be [pnpm](https://pnpm.io/).
+The TutorialKit repo is a monorepo using pnpm workspaces. The package manager used to install and link dependencies must be [pnpm](https://pnpm.io/). pnpm version 8.15.6 is required (npm i -g pnpm@8.15.6)
```
AriPerkkio, line 7: Instead of hard-coding the version here, we should mention that we use `packageManager` field and utilize Node's Corepack.
 
```suggestion
The TutorialKit repo is a monorepo using pnpm workspaces. The package manager used to install and link dependencies must be [pnpm](https://pnpm.io/). Package manager versioning is handled by [`packageManager`](https://nodejs.org/api/packages.html#packagemanager) field that's supported by [Node's Corepack](https://nodejs.org/api/corepack.html) and other tools.
```

noam-honig:
@AriPerkkio Added your change and also added where one can look for the specific verison - do you think its required?

AriPerkkio:
> Added your change and also added where one can look for the specific verison - do you think its required?

Corepack can automatically detect and use the correct version of defined package manager, so there's no need for developers to check what version is there. I think it's better to delete that part from the docs.

AriPerkkio (approved):

noam-honig:
@AriPerkkio Thanks - sorry, I wasn't available for a few hours to remove the line you wanted

Issue #263, opened 2024-08-19
Allow the user to add files


AriPerkkio:
### Discussed in https://github.com/stackblitz/tutorialkit/discussions/204

<div type='discussions-op-text'>

<sup>Originally posted by **GaetanRdn** August  1, 2024</sup>
> Like in Stackblitz, the possibility to add file from the file tree could be a great feature.
> 
> We could create bigger exercices, more difficult. 
> 
> In my case, I am preparing a tutorial for Angular. And with it, we quickly need to create new files.
</div>

This feature request is about adding new files using the file tree. This is something we can implement right away.

In https://github.com/stackblitz/tutorialkit/issues/208 there is a feature request for adding files via terminal.

Issue #208, opened 2024-08-02
Update _files content with scrips command


geromegrignon:
### Is your feature request related to a problem?

I'm creating a NgRx (Angular state management library) and the first step is to add the library with the `ng add @ngrx/store@latest`. It adds the dependency in package.json and updates some files.

I added all these files in the _files folder as the student task is to run the script and sees how it affect the project.
However even if the command succeeds, the files are not updated.

### Describe the solution you'd like.

_files content should be updated from terminal action.

### Describe alternatives you've considered.

No alternative

GaetanRdn:
The solution for this problem could be the same for https://github.com/stackblitz/tutorialkit/discussions/204 in a first time.

PR #314, opened 2024-09-04
feat: add files via file tree

AriPerkkio:
- Closes https://github.com/stackblitz/tutorialkit/issues/263
- Related to https://github.com/stackblitz/tutorialkit/discussions/204
- Not related to https://github.com/stackblitz/tutorialkit/issues/208. That one requires WebContainer changes so that we can listen for fs changes.
- While working on the `FileTree` I noticed plenty of issues with it and created https://github.com/stackblitz/tutorialkit/issues/326 as follow-up. 

### BREAKING CHANGES

- Projects that are using `@tutorialkit/react` package directly, **without TutorialKit** need to update `<FileTree>` component's `props.files` type

### Description

Adds new metadata option that can be used to enable file and folder creation in the file tree. By default this feature is disabled.

```yaml
---
type: lesson
editor:
  fileTree:
    allowEdits: true
---
```

[tk-filetree-example.webm](https://github.com/user-attachments/assets/d9a664ee-1af6-4cf4-9f20-addbaa3fb381)

### DIFF
--- a/docs/tutorialkit.dev/src/components/react-examples/ExampleFileTree.tsx
+++ b/docs/tutorialkit.dev/src/components/react-examples/ExampleFileTree.tsx
@@ -14,7 +14,7 @@ export default function ExampleFileTree() {
       selectedFile={selectedFile}
       onFileSelect={setSelectedFile}
       onFileChange={(event) => {
-        if (event.method === 'ADD') {
+        if (event.method === 'add') {
           setFiles([...files, { path: event.value, type: event.type }].sort());
         }
       }}
--- a/docs/tutorialkit.dev/src/components/react-examples/ExampleFileTree.tsx
+++ b/docs/tutorialkit.dev/src/components/react-examples/ExampleFileTree.tsx
@@ -15,7 +15,7 @@ export default function ExampleFileTree() {
       onFileSelect={setSelectedFile}
       onFileChange={(event) => {
         if (event.method === 'add') {
-          setFiles([...files, { path: event.value, type: event.type }].sort());
+          setFiles([...files, { path: event.value, type: event.type }]);
         }
       }}
     />
--- a/docs/tutorialkit.dev/src/components/react-examples/ExampleFileTree.tsx
+++ b/docs/tutorialkit.dev/src/components/react-examples/ExampleFileTree.tsx
@@ -2,25 +2,31 @@ import { useState } from 'react';
 import FileTree from '@tutorialkit/react/core/FileTree';
 
 export default function ExampleFileTree() {
-  const [selectedFile, setSelectedFile] = useState(FILES[0]);
+  const [files, setFiles] = useState(INITIAL_FILES);
+  const [selectedFile, setSelectedFile] = useState(INITIAL_FILES[0]);
 
   return (
     <FileTree
-      files={FILES}
+      files={files}
       hideRoot
       className="my-file-tree"
       hiddenFiles={['package-lock.json']}
       selectedFile={selectedFile}
       onFileSelect={setSelectedFile}
+      onFileChange={(event) => {
+        if (event.method === 'ADD') {
+          setFiles([...files, event.value].sort());
+        }
+      }}
     />
   );
 }
 
-const FILES = [
-  '/src/index.js',
-  '/src/index.html',
-  '/src/assets/logo.svg',
+const INITIAL_FILES = [
   '/package-lock.json',
   '/package.json',
+  '/src/assets/logo.svg',
+  '/src/index.html',
+  '/src/index.js',
   '/vite.config.js',
 ];
--- a/docs/tutorialkit.dev/src/components/react-examples/ExampleFileTree.tsx
+++ b/docs/tutorialkit.dev/src/components/react-examples/ExampleFileTree.tsx
@@ -1,9 +1,9 @@
-import { useState } from 'react';
+import { useState, type ComponentProps } from 'react';
 import FileTree from '@tutorialkit/react/core/FileTree';
 
 export default function ExampleFileTree() {
   const [files, setFiles] = useState(INITIAL_FILES);
-  const [selectedFile, setSelectedFile] = useState(INITIAL_FILES[0]);
+  const [selectedFile, setSelectedFile] = useState(INITIAL_FILES[0].path);
 
   return (
     <FileTree
@@ -15,18 +15,18 @@ export default function ExampleFileTree() {
       onFileSelect={setSelectedFile}
       onFileChange={(event) => {
         if (event.method === 'ADD') {
-          setFiles([...files, event.value].sort());
+          setFiles([...files, { path: event.value, type: event.type }].sort());
         }
       }}
     />
   );
 }
 
-const INITIAL_FILES = [
-  '/package-lock.json',
-  '/package.json',
-  '/src/assets/logo.svg',
-  '/src/index.html',
-  '/src/index.js',
-  '/vite.config.js',
+const INITIAL_FILES: ComponentProps<typeof FileTree>['files'] = [
+  { path: '/package-lock.json', type: 'FILE' },
+  { path: '/package.json', type: 'FILE' },
+  { path: '/src/assets/logo.svg', type: 'FILE' },
+  { path: '/src/index.html', type: 'FILE' },
+  { path: '/src/index.js', type: 'FILE' },
+  { path: '/vite.config.js', type: 'FILE' },
 ];
--- a/docs/tutorialkit.dev/src/components/react-examples/ExampleFileTree.tsx
+++ b/docs/tutorialkit.dev/src/components/react-examples/ExampleFileTree.tsx
@@ -23,10 +23,10 @@ export default function ExampleFileTree() {
 }
 
 const INITIAL_FILES: ComponentProps<typeof FileTree>['files'] = [
-  { path: '/package-lock.json', type: 'FILE' },
-  { path: '/package.json', type: 'FILE' },
-  { path: '/src/assets/logo.svg', type: 'FILE' },
-  { path: '/src/index.html', type: 'FILE' },
-  { path: '/src/index.js', type: 'FILE' },
-  { path: '/vite.config.js', type: 'FILE' },
+  { path: '/package-lock.json', type: 'file' },
+  { path: '/package.json', type: 'file' },
+  { path: '/src/assets/logo.svg', type: 'file' },
+  { path: '/src/index.html', type: 'file' },
+  { path: '/src/index.js', type: 'file' },
+  { path: '/vite.config.js', type: 'file' },
 ];
--- a/packages/react/src/core/ContextMenu.tsx
+++ b/packages/react/src/core/ContextMenu.tsx
@@ -1,15 +1,15 @@
 import { useRef, useState, type ComponentProps } from 'react';
 import { Root, Portal, Content, Item, Trigger } from '@radix-ui/react-context-menu';
-import type { I18n } from '@tutorialkit/types';
+import type { FileDescriptor, I18n } from '@tutorialkit/types';
 
 interface FileChangeEvent {
-  type: 'FILE' | 'FOLDER';
-  method: 'ADD' | 'REMOVE' | 'RENAME';
+  type: FileDescriptor['type'];
+  method: 'add' | 'remove' | 'rename';
   value: string;
 }
 
 interface FileRenameEvent extends FileChangeEvent {
-  method: 'RENAME';
+  method: 'rename';
   oldValue: string;
 }
 
@@ -39,7 +39,7 @@ export function ContextMenu({
   triggerProps,
   ...props
 }: Props) {
-  const [state, setState] = useState<'IDLE' | 'ADD_FILE' | 'ADD_FOLDER'>('IDLE');
+  const [state, setState] = useState<'idle' | 'add_file' | 'add_folder'>('idle');
   const inputRef = useRef<HTMLInputElement>(null);
 
   if (!onFileChange) {
@@ -52,12 +52,12 @@ export function ContextMenu({
     if (name) {
       onFileChange?.({
         value: `${directory}/${name}`,
-        type: state === 'ADD_FILE' ? 'FILE' : 'FOLDER',
-        method: 'ADD',
+        type: state === 'add_file' ? 'file' : 'folder',
+        method: 'add',
       });
     }
 
-    setState('IDLE');
+    setState('idle');
   }
 
   function onFileNameKeyPress(event: React.KeyboardEvent<HTMLInputElement>) {
@@ -67,7 +67,7 @@ export function ContextMenu({
   }
 
   function onCloseAutoFocus(event: Event) {
-    if ((state === 'ADD_FILE' || state === 'ADD_FOLDER') && inputRef.current) {
+    if ((state === 'add_file' || state === 'add_folder') && inputRef.current) {
       event.preventDefault();
       inputRef.current.focus();
     }
@@ -83,9 +83,9 @@ export function ContextMenu({
     <Root>
       {position === 'before' && element}
 
-      {state !== 'IDLE' && (
+      {state !== 'idle' && (
         <div className="flex items-center gap-2 border-2 border-solid border-transparent" {...props}>
-          <div className={`scale-120 shrink-0 ${state === 'ADD_FILE' ? 'i-ph-file-duotone' : 'i-ph-folder-duotone'}`} />
+          <div className={`scale-120 shrink-0 ${state === 'add_file' ? 'i-ph-file-duotone' : 'i-ph-folder-duotone'}`} />
           <input
             ref={inputRef}
             autoFocus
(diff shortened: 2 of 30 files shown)


AriPerkkio:
Early draft PR, mostly so that 81eac67d57180f7df4b6bb699dddb21fa58ee519 can be used if #208 is started before this is finalized.

packages/react/src/core/ContextMenu.tsx
```diff
@@ -0,0 +1,132 @@
+import { useRef, useState, type ComponentProps } from 'react';
+import { Root, Portal, Content, Item, Trigger } from '@radix-ui/react-context-menu';
+import type { I18n } from '@tutorialkit/types';
+
+interface FileChangeEvent {
+  type: 'FILE' | 'FOLDER';
+  method: 'ADD' | 'REMOVE' | 'RENAME';
```
AriPerkkio, line 7: Added the `REMOVE` and `RENAME` here even though they are not yet implemented. This kind of API allows us to extend the functionality later without causing breaking changes.

packages/types/src/schemas/common.ts
```diff
@@ -161,8 +161,32 @@ export const terminalSchema = z.union([
...
+
+  z.strictObject({
+    fileTree: z
+      .union([
+        // or you can only remove the file tree
+        z.boolean(),
+
+        // or configure file tree with options
+        z.strictObject({
+          allowEdits: z
```
Nemikolh, line 176: I think we want something more fine grained where you specify in which scope you want to allow files and folder to be added.

packages/runtime/src/store/index.ts
```diff
@@ -304,6 +310,27 @@ export class TutorialStore {
...
 
+  addFile(filePath: string) {
+    // prevent creating duplicates
+    if (this._editorStore.files.get().includes(filePath)) {
+      return this.setSelectedFile(filePath);
+    }
+
+    this._editorStore.addFileOrFolder(filePath);
+    this.setSelectedFile(filePath);
+    this._runner.updateFile(filePath, '');
```
Nemikolh, line 321: Nit:

```suggestion
    // always select the existing or newly created file
    this.setSelectedFile(filePath);

    // prevent duplicates
    if (this._editorStore.files.get().includes(filePath)) {
      return;
    }

    this._editorStore.addFileOrFolder(filePath);
    this._runner.updateFile(filePath, '');
```

packages/runtime/src/store/index.ts
```diff
@@ -239,19 +249,15 @@ export class TutorialStore {
...
 
   hasEditor(): boolean {
     if (!this._lesson) {
       return false;
     }
 
-    const { editor } = this._lesson.data;
-
-    return editor !== false;
+    return this.editorConfig.get().visible;
```
Nemikolh, line 260: Nice improvement! :pinched_fingers:

packages/runtime/src/store/editor.ts
```diff
@@ -13,12 +14,14 @@ export interface ScrollPosition {
...
 
-export type EditorDocuments = Record<string, EditorDocument>;
+export type EditorDocuments = Record<string, EditorDocument | undefined>;
 
 export class EditorStore {
+  editorConfig = atom<EditorConfig>(new EditorConfig());
   selectedFile = atom<string | undefined>();
   documents = map<EditorDocuments>({});
 
+  files = computed(this.documents, (documents) => Object.keys(documents).sort());
```
Nemikolh, line 24: We should provide a custom sort that always sort folders first which is what we get in vscode:

```suggestion
  files = computed(this.documents, (documents) => Object.keys(documents).sort((a, b) => {
    const segmentsA = a.split('/');
    const segmentsB = b.split('/');

    const minLength = Math.min(segmentsA.length, segmentsB.length);

    // order parent folder by name
    for (let i = 0; i < minLength - 1; i++) {
      const comparison = compareString(segmentsA[i], segmentsB[i]);

      if (comparison !== 0) {
        return comparison;
      }
    }

    // if they have a different length, folder get shown first
    if (segmentsA.length > minLength) {
      return -1;
    }

    if (segmentsB.length > minLength) {
      return 1;
    }

    return compareString(segmentsA[minLength - 1], segmentsB[minLength - 1]);
  });
```

where `compareString` is simply defined as:

```ts
function compareString(a: string, b: string) {
  if (a < b) {
    return -1;
  }

  if (a > b) {
    return 1;
  }

  return 0;
}
```

packages/react/src/core/FileTree.tsx
```diff
@@ -231,7 +259,7 @@ function buildFileList(
 
       if (depth === segments.length - 1) {
         fileList.push({
-          kind: 'file',
+          kind: name.includes('.') ? 'file' : 'folder',
```
Nemikolh, line 262: Oh this is a bit hacky. I can see why you needed something like this given that now `files` no longer contains "just" files but can also contains folders. I think we should change the `files` prop from `string[]`  to `{ path: string, type: 'file' | 'folder' }[]`

packages/react/src/core/FileTree.tsx
```diff
@@ -262,3 +291,7 @@ function isHiddenFile(filePath: string, fileName: string, hiddenFiles: Array<str
     return pathOrRegex.test(filePath);
   });
 }
+
+function getDepthStyle(depth: number) {
+  return { paddingLeft: `${12 + depth * NODE_PADDING_LEFT}px` };
```
Nemikolh, line 296: I hope someday we can make that a bit better. At a minimum using `rem` instead of `px` and then maybe using a `calc` and a CSS variable to make it possible to tweak the indentation.

packages/react/src/core/FileTree.tsx
```diff
@@ -73,7 +86,7 @@ export function FileTree({ files, onFileSelect, selectedFile, hideRoot, scope, h
   }
 
   return (
-    <div className={classNames(className, 'transition-theme bg-tk-elements-fileTree-backgroundColor')}>
+    <div className={classNames(className, 'h-100% transition-theme bg-tk-elements-fileTree-backgroundColor')}>
```
Nemikolh, line 89:
```suggestion
    <div className={classNames(className, 'h-full transition-theme bg-tk-elements-fileTree-backgroundColor')}>
```

packages/react/src/core/FileTree.tsx
```diff
@@ -91,13 +104,24 @@ export function FileTree({ files, onFileSelect, selectedFile, hideRoot, scope, h
...
         }
       })}
+
+      <ContextMenu // context menu for the remaining free space area
+        position="after"
+        i18n={i18n}
+        style={getDepthStyle(0)}
+        directory=""
+        onFileChange={onFileChange}
+        triggerProps={{ className: 'h-100%', 'data-testid': 'file-tree-root-context-menu' }}
```
Nemikolh, line 123:
```suggestion
        triggerProps={{ className: 'h-full', 'data-testid': 'file-tree-root-context-menu' }}
```

packages/react/src/core/ContextMenu.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  if (!onFileChange) {
+    return children;
+  }
+
+  function onFileNameEnd(event: React.KeyboardEvent<HTMLInputElement> | React.FocusEvent<HTMLInputElement>) {
+    const name = event.currentTarget.value;
+    const isFile = state === 'ADD_FILE';
+
+    // files must contain extension
```
Nemikolh, line 53: I don't think we should not make that assumption. Extensions are not mandatory anywhere in tutorialkit AFAIK.

Nemikolh (commented):
Did a first pass, good work on this!! :muscle:

packages/react/src/core/FileTree.tsx
AriPerkkio: I also thought about this at first but didn't want to break existing integrations. But I'll mark this as `breaking change` and change the prop type.

packages/react/src/core/FileTree.tsx
AriPerkkio: This might get fixed during https://github.com/stackblitz/tutorialkit/issues/326 automatically. But yep I agree on all points.

packages/runtime/src/store/editor.ts
AriPerkkio: Yup sounds good! Though for this to work, we'll need to be able to identify folders from files. Right now this sorting function get confused about empty folders:

[tk-filetree-sort.webm](https://github.com/user-attachments/assets/2087e758-e2da-4442-8ccf-25aee658b407)

packages/types/src/schemas/common.ts
AriPerkkio: Sounds good! Though this is something that could be implemented after this first PR without breaking the API too, so that `true` would mean "allow everything". 

But I can also take a look at this once I've addressed other review findings. 👍

packages/react/src/core/ContextMenu.tsx
AriPerkkio: I agree - we should allow files without extensions. Need https://github.com/stackblitz/tutorialkit/pull/314#discussion_r1753809729 for that. 👷

packages/runtime/src/store/editor.ts
Nemikolh: Oh good point, when I wrote that, I forgot we now have empty folders

AriPerkkio: This ended up being a bit more complicated now that files and folders can be mixed there but I think I got it! Also added bunch of unit tests for this.

AriPerkkio:
@Nemikolh ready for review round 2! 🔔  

Added support for identifying files and folders from each other - we no longer require files to have extensions.

However there's still an edge case that causes issues. As we are using filepath as map key here, there's no way for us to have file and folder with same name in the tree. 

https://github.com/stackblitz/tutorialkit/blob/38df1f39f16ff3c7f228cd90c4ecdf2a71232072/packages/runtime/src/store/editor.ts#L16

So for example this will fail:

```ts
test('empty directories are removed when new content is added', () => {
  const store = new EditorStore();
  store.setDocuments({
    example: 'should not be confused as directory',
  });

  store.addFileOrFolder({ path: 'example', type: 'FOLDER' });
  store.addFileOrFolder({ path: 'example/another', type: 'FOLDER' });

  expect(store.files.get().map(toFilename)).toMatchInlineSnapshot(`
    [
      "example/another",
      "example",
    ]
  `);
});
```

Is this really an issue that users could run into? Maybe not.

Nemikolh:
@AriPerkkio I'm not up to date with every file systems that exists out there, but AFAIK, in most of them you cannot have a file and a folder named the same. If you think about it, that make sense with the Linux [`open`](https://man7.org/linux/man-pages/man2/open.2.html) function and its equivalent in node.

In our case, what matter is what the WebContainer fs does and it does not allow multiple files given a single path. So this is completely fine :+1:

Nemikolh (commented):
Second pass done! I hope I didn't miss anything. I'll probably do another pass tomorrow :pray:

packages/runtime/src/store/editor.ts
```diff
@@ -158,3 +169,46 @@ export class EditorStore {
...
+
+    const comparison = compareString(a.path, b.path);
+
+    // either folder name changed or last segments are compared
+    if (comparison !== 0 || a.isLast || b.isLast) {
+      return comparison;
+    }
+  }
+
+  throw new Error(JSON.stringify({ fileA, fileB }));
```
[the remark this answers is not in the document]
AriPerkkio: Oops, this was for debugging. I'll add the applied change!

packages/react/src/core/FileTree.tsx
```diff
@@ -215,18 +216,19 @@ interface BaseNode {
...
 }
 
 interface FileNode extends BaseNode {
-  kind: 'file';
+  kind: 'FILE';
 }
 
 interface FolderNode extends BaseNode {
-  kind: 'folder';
+  kind: 'FOLDER';
```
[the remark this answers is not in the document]
AriPerkkio: No strong opinions, I'll change to lowercase.

packages/react/src/core/ContextMenu.tsx
```diff
@@ -0,0 +1,130 @@
...
+export function ContextMenu({
+  onFileChange,
+  directory,
+  i18n,
+  position = 'before',
+  children,
+  triggerProps,
+  ...props
+}: Props) {
+  const [state, setState] = useState<'IDLE' | 'ADD_FILE' | 'ADD_FOLDER'>('IDLE');
```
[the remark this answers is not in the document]
AriPerkkio: Not really, hated it in Redux 🙃 

I'm typically using uppercases in string literals, in cases where they are used enum-like. Not sure why, but somewhere I've got this convention. 🤔 
I can change these to lowercase as well.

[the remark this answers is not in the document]
Nemikolh: Oh LOL, I had completely forgotten about that convention! I haven't done redux for nearly a decade now :joy:

docs/tutorialkit.dev/src/components/react-examples/ExampleFileTree.tsx
```diff
@@ -1,26 +1,32 @@
...
-      files={FILES}
+      files={files}
       hideRoot
       className="my-file-tree"
       hiddenFiles={['package-lock.json']}
       selectedFile={selectedFile}
       onFileSelect={setSelectedFile}
+      onFileChange={(event) => {
+        if (event.method === 'ADD') {
+          setFiles([...files, { path: event.value, type: event.type }].sort());
```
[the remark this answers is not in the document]
AriPerkkio: I like the idea! Currently the functionality is in `runtime` though - I'll see if moving it from there to component looks good.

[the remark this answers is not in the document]
Nemikolh: Oh I forgot that it's in the `runtime`, maybe it can be imported by the `FileTree` and re-exported as a static?

packages/types/src/entities/index.ts
```diff
@@ -3,6 +3,7 @@ import type { ChapterSchema, LessonSchema, PartSchema } from '../schemas/index.j
 
 export type * from './nav.js';
 
+export type File = { path: string; type: 'FILE' | 'FOLDER' };
```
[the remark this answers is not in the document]
AriPerkkio: `FileDescriptor` sounds good. I didn't like the `File` either.

docs/tutorialkit.dev/src/components/react-examples/ExampleFileTree.tsx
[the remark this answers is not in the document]
AriPerkkio: We don't want to add `runtime` as dependency for `react` package though. 🤔

packages/react/src/core/ContextMenu.tsx
```diff
@@ -0,0 +1,130 @@
...
+  onFileChange?: (event: FileChangeEvent | FileRenameEvent) => void;
+
+  /** Directory of the clicked file. */
+  directory: string;
+
+  /** Whether to render new files/directories before or after the trigger element. Defaults to `'before'`. */
+  position?: 'before' | 'after';
+
+  /** Localized texts for menu. */
+  i18n?: Pick<I18n, 'fileTreeCreateFileText' | 'fileTreeCreateFolderText'>;
```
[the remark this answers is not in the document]
AriPerkkio: Yep we should! I saw that there's quite much duplication in couple of places. When we can derive props from their orignal source, we should always do it.

packages/runtime/src/store/index.ts
```diff
@@ -311,23 +311,25 @@ export class TutorialStore {
   }
 
   addFile(filePath: string) {
+    // always select the existing or newly created file
+    this.setSelectedFile(filePath);
+
     // prevent creating duplicates
-    if (this._editorStore.files.get().includes(filePath)) {
-      return this.setSelectedFile(filePath);
+    if (this._editorStore.files.get().find((file) => file.type === 'FILE' && file.path === filePath)) {
```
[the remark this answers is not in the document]
AriPerkkio: Good catch! No need to check for type here.

docs/tutorialkit.dev/src/components/react-examples/ExampleFileTree.tsx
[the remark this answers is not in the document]
Nemikolh: It's already one:

https://github.com/stackblitz/tutorialkit/blob/38df1f39f16ff3c7f228cd90c4ecdf2a71232072/packages/react/package.json#L78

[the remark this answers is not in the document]
AriPerkkio: Oh right, it is. It should actually be just development time dependency, as we are only utilizing types from there. But this change would require it in runtime too.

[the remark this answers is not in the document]
AriPerkkio: Maybe we don't even need to sort the files in the editor. It's enough to sort them in the file tree only. Let's see... 👀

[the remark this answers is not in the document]
Nemikolh: Oh yeah we could also simply sort them there.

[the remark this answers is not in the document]
AriPerkkio: Alright, moved the sorting directly to the component only. I think it makes sense, as order matters only when we present the files in tree visually. What do you think?

Nemikolh (approved):
Incredible job on this @AriPerkkio! :partying_face: :partying_face: 

I'm starting to think we could merge this PR first, and then change the API in a follow-up PR to be:

```ts
type GlobPattern = string;

interface P {
  allowEdits?: boolean | GlobPattern | GlobPattern[]
}
```

That way it wouldn't be a breaking change, we would still support the boolean version with the default value being `false`.

I got a feeling that some will still want to use `true` anyway. 

What I think we should do in the follow-up PR is to clarify in the docs that only allowing certain patterns (or even a single path) should be preferred.

I would even reference the talk from Rich Harris: FullStack Documentation as a suggestion to why this should be preferred.

AriPerkkio:
I agree on all points! I'll continue on that next.

PR #318, opened 2024-09-05
fix(react): stale lesson data after navigation

AriPerkkio:
- Fixes issues where stale lesson data (metadata) is used after navigating to new lesson
- Deployed at https://github.com/AriPerkkio/tutorial-vite-plugin/pull/24

In https://github.com/stackblitz/tutorialkit/pull/253 the amount of "unnecessary" re-renders were reduced by moving state to components where it was actually needed. 

It turns out, those re-renders were indeed needed as `<WorkspacePanel />` is accessing non-reactive references of the `tutorialStore` instead of reactive ones. This is not easily visible from the code though. We should improve that part later. This is now just a quick fix for the high priority bug.

These are the non-reactive methods of `tutorialStore` that use stale data:

https://github.com/stackblitz/tutorialkit/blob/1eb41fd917ceff08fe9484bd07e732cb673a7a1f/packages/react/src/Panels/WorkspacePanel.tsx#L35-L37

Failing test case on CI: https://github.com/stackblitz/tutorialkit/actions/runs/10733493214/job/29766948087?pr=318

Before:

[tk-layout-after-nav-bug.webm](https://github.com/user-attachments/assets/be25e20f-6805-45fe-8baa-be50b3de9812)

After:

[tk-layout-after-nav-fixed.webm](https://github.com/user-attachments/assets/31016e42-4f91-4b21-a9c1-9cf4435df3d3)

### DIFF
--- a/packages/react/src/Panels/WorkspacePanel.tsx
+++ b/packages/react/src/Panels/WorkspacePanel.tsx
@@ -32,6 +32,13 @@ interface TerminalProps extends PanelProps {
  * This component is the orchestrator between various interactive components.
  */
 export function WorkspacePanel({ tutorialStore, theme }: Props) {
+  /**
+   * Re-render when lesson changes.
+   * The `tutorialStore.hasEditor()` and other methods below access
+   * stale data as they are not reactive.
+   */
+  useStore(tutorialStore.ref);
+
   const hasEditor = tutorialStore.hasEditor();
   const hasPreviews = tutorialStore.hasPreviews();
   const hideTerminalPanel = !tutorialStore.hasTerminalPanel();
@@ -89,6 +96,7 @@ function EditorSection({ theme, tutorialStore, hasEditor }: PanelProps) {
   const selectedFile = useStore(tutorialStore.selectedFile);
   const currentDocument = useStore(tutorialStore.currentDocument);
   const lessonFullyLoaded = useStore(tutorialStore.lessonFullyLoaded);
+  const storeRef = useStore(tutorialStore.ref);
 
   const lesson = tutorialStore.lesson!;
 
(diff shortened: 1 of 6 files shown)


packages/react/src/Panels/WorkspacePanel.tsx
```diff
@@ -32,6 +32,13 @@ interface TerminalProps extends PanelProps {
...
+  /**
+   * Re-render when lesson changes.
+   * The `tutorialStore.hasEditor()` and other methods below access
+   * stale data as they are not reactive.
+   */
+  useStore(tutorialStore.ref);
+
   const hasEditor = tutorialStore.hasEditor();
   const hasPreviews = tutorialStore.hasPreviews();
   const hideTerminalPanel = !tutorialStore.hasTerminalPanel();
```
AriPerkkio, line 44: Ideally we should make these reactive instead, e.g.

```ts
const hasEditor = useStore(tutorialStore.hasEditor);
```

... but as that would require quite much refactoring on the store, e.g. making `lesson` reactive, I'd leave that out side of this PR.

AriPerkkio:
@Nemikolh we have a failing minimal test case now! 🎉 🎉 

After yesterday's extensive debugging session I realized what the root cause actually was. The `WorkspacePanel` is accessing non-reactive references of store here:

https://github.com/stackblitz/tutorialkit/blob/1eb41fd917ceff08fe9484bd07e732cb673a7a1f/packages/react/src/Panels/WorkspacePanel.tsx#L35-L37

And as `WorkspacePanel` is not subscribing to any other changes (as it doesn't really need to), it won't re-render when parts of the store change.

I tried to make `lesson` reactive like we tried yesterday, but ran into plenty of weird React hydration mismatch errors afterwards. As this bug is critical, I would suggest that we merge this with the work-around and publish new release with the fix.

Nemikolh:
Good work on the test! 🤩 

Sounds good for the rest! I will review this ASAP. 👍

Nemikolh (approved):
Looks good!

PR #325, opened 2024-09-11
chore: uno css improvements for local development

AriPerkkio:
While working on https://github.com/stackblitz/tutorialkit/pull/314 noticed that when adding new UnoCSS classes on `packages/react`, these classes would not be auto reloaded in `packages/template` development mode. Adding `content.pipeline` seems to fix this.

Also UnoCSS VSCode extension requires uno config to be present - let's add one in root of the monorepo.


uno.config.js
Nemikolh, line 1: Nice! I've been using an uncommited config for just that :sweat_smile: 

Now that we have the `defineConfig` from the theme package, this is pretty nice :sunglasses:

packages/template/uno.config.ts
```diff
@@ -1,5 +1,10 @@
 import { defineConfig } from '@tutorialkit/theme';
 
 export default defineConfig({
-  // add your UnoCSS config here: https://unocss.dev/guide/config-file
+  // required for TutorialKit monorepo development mode
+  content: {
+    pipeline: {
+      include: '**',
+    },
+  },
```
Nemikolh, line 9: Really cool trick! I never managed to get pipeline working, glad you found a way!

Nemikolh (approved):
Super nice DX improvement here! Good job @AriPerkkio ! :raised_hands:

PR #328, opened 2024-09-11
introduce an ImageContent component in editor

### LINKED ISSUE #309: Display images

Codemirror doesn't seem to support images
image

Display images within the code editor

This behaviour is implemented in the react package CodeMirrorEditor: "packages/react/src/core/CodeMirrorEditor/BinaryContent.tsx" which is triggered there:
"packages/react/src/core/CodeMirrorEditor/index.tsx"

### Additional context

I think this is a good first issue:
- we can detect images via mime-type
- and then display them
Would a PR creating a new "ImageContent" component be accepted? Does display images create any sort of security risk?

eric-burel:
This PR introduces an ImageContent component for the editor, for binary images.

Design choices:
- only binary images are supported, with format you typically expect on the web
- svg is still treated as text (you might want a tutorial about programming with SVG)
- other binary image format are not supported (which I think is ok for most use cases)

Problems left: 
- somehow the editor displays code from another file alongside the image
- not sure if shoving images in the first demo lesson is good for testing

image

Closes #309


packages/react/src/core/CodeMirrorEditor/ImageContent.tsx
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  const ext = split[split.length - 1];
+
+  return binaryImageExtensions.has(ext);
+}
+
+export function ImageContent({ src }: { src: string }) {
+  return <img src={src} />;
+}
```
AriPerkkio, line 24: > Problems left:
> somehow the editor displays code from another file alongside the image

The existing `<BinaryContent />` does too, but it uses `position: absolute; inset: 0` to cover the text area. We could do same here:

```suggestion
export function ImageContent({ src }: { src: string }) {
  return (
    <div className="flex items-center justify-center absolute inset-0 z-10 bg-tk-elements-app-backgroundColor">
      <img src={src} />
    </div>
  );
}
```

This isn't actually good approach as the text content of previous file is still present in DOM. It's part of tab and read order so users can actually navigate to this invisible element. But as this bug is already present in TutorialKit we can ignore it on this PR. 🤷‍♂️

Nemikolh: I think this is a leftover of a time where we were trying to cache the `EditorView` from CodeMirror which I don't think is necessary. We should change the logic so that the `CodeMirrorEditor` is not rendered in that case.

PR #332, opened 2024-09-16
feat: support glob patterns in `editor.fileTree.allowEdits`

AriPerkkio:
- Continues https://github.com/stackblitz/tutorialkit/pull/314

Adds support for users to specify one or multiple glob patterns in `editor.fileTree.allowEdits`. 

```yaml
---
type: lesson
title: Example
editor:
  fileTree:
    allowEdits:
      # Items in root
      - "/*"

      # Only "allowed-filename-only.js" inside "/first-level" folder
      - "/first-level/allowed-filename-only.js"

      # Anything inside "second-level" folders anywhere
      - "**/second-level/**"
---
```

When students attempt to create file that doesn't match these patterns, we show an error dialog:


packages/react/src/core/ContextMenu.tsx
```diff
@@ -1,6 +1,8 @@
-import { useRef, useState, type ComponentProps } from 'react';
+import { useRef, useState, type ComponentProps, type ReactNode } from 'react';
 import { Root, Portal, Content, Item, Trigger } from '@radix-ui/react-context-menu';
-import type { FileDescriptor, I18n } from '@tutorialkit/types';
+import * as RadixDialog from '@radix-ui/react-dialog';
+import picomatch from 'picomatch/posix';
```
AriPerkkio, line 4: This entrypoint is compatible with browser/non-Node environments.

> #### Example without node.js
> 
> For environments without `node.js`, `picomatch/posix` provides you a dependency-free matcher, without automatic OS detection.
> https://github.com/micromatch/picomatch?tab=readme-ov-file#api

packages/react/src/core/ContextMenu.tsx
```diff
@@ -17,44 +19,80 @@ interface Props extends ComponentProps<'div'> {
...
   triggerProps?: ComponentProps<'div'> & { 'data-testid'?: string };
 }
 
+const i18nDefaults = {
+  fileTreeFailedToCreateFileText: 'Failed to create file "${filename}".',
+  fileTreeFailedToCreateFolderText: 'Failed to create folder "${filename}".',
+  fileTreeCreateFileText: 'Create file',
+  fileTreeCreateFolderText: 'Create folder',
+  fileTreeAllowedPatternsText: 'Allowed patterns are:',
+} as const satisfies Props['i18n'];
```
AriPerkkio, line 51: As React components can be used without TutorialKit, we need to store copies of localizations in these components as well. We might want to move those default localizations in separate package, or serve defaults from `@tutorialkit/types` maybe. 🤔

packages/react/src/core/ContextMenu.tsx
```diff
@@ -128,3 +183,25 @@ function MenuItem({ icon, children, ...props }: { icon: string } & ComponentProp
...
+
+            {children}
+
+            <RadixDialog.Close title="Close" className="absolute top-4 right-4 w-6 h-6">
+              <span aria-hidden className="i-ph-x block w-full h-full"></span>
+            </RadixDialog.Close>
+          </div>
+        </RadixDialog.Content>
+      </RadixDialog.Portal>
+    </RadixDialog.Root>
```
AriPerkkio, line 205: @sulco, @Nemikolh, I wasn't really sure about what to put here and what kind of styles to use. Feel free to propose any changes here. Right now It's trying to follow TK's other styles, e.g. by using `bg-tk-background-negative`. 

In future (before next release) we will also show here error when user tries to create files that already exist on file system, e.g. creating `vite.config.ts` that's on template already.

packages/react/src/core/ContextMenu.tsx
Nemikolh: Oh that's a good point, I think having them in `@tutorialkit/types` make sense.

packages/react/src/core/ContextMenu.tsx
Nemikolh: A few thoughts:
* We could probably follow what Svelte's tutorial is doing, aka:
  * Replace `Error` with `Cannot create this file` or `This action is not allowed`
  * Have an `Ok` button
  * Maybe let's use the app background color instead of the red which I find too strong
* I wonder if we should make that component overridable from the start, however a future PR is also fine

For reference this is what Svelte dialog looks like:

AriPerkkio: Alright, now using following. What do you think?

And once I add the _"prevent creating files that exists on fs already, e.g. in templates"_ fix, we would show here `"This action is not allowed, File exists already"`.

image

image

AriPerkkio: > I wonder if we should make that component overridable from the start, however a future PR is also fine

I agree this would be useful. But I'm not yet sure how this could be implemented, as this component is so deeply nested the `@tutorialkit/react` components. For override components, we need the logic to be in `@tutorialkit/astro` level. And previously override components could even be `.astro` files. 🤔

packages/react/src/Panels/WorkspacePanel.tsx
```diff
@@ -161,6 +161,7 @@ function EditorSection({ theme, tutorialStore, hasEditor }: PanelProps) {
         onHelpClick={lessonFullyLoaded ? onHelpClick : undefined}
         onFileSelect={(filePath) => tutorialStore.setSelectedFile(filePath)}
         onFileTreeChange={editorConfig.fileTree.allowEdits ? onFileTreeChange : undefined}
```
Nemikolh, line 163: Oh I see! If the callback is not provided then we don't allow any edits. I'm not sure I'm a fan of this approach, my instinct would have been that if `allowEditPatterns` is `undefined` then nothing is allowed as opposed to everything.

I would even go further and say that all those should have the same behaviour:

```ts
allowEditPatterns: [],
allowEditPatterns: undefined,
allowEditPatterns: [!**],
```

Nemikolh (commented):
Really good work on this! :star_struck: 

Some very minor things, and let's ship this! :muscle:

packages/react/src/core/ContextMenu.tsx
Nemikolh: I think we could provide the component via a React Context (or just do some form of dynamic linking). It would have the inconvenient that the default component is always included in the bundle but that's probably fine?

docs/tutorialkit.dev/src/content/docs/reference/react-components.mdx
```diff
@@ -116,6 +116,8 @@ A component to list files in a tree view.
   }
   ```
 
+* `allowEditPatterns?: string[]` - Glob patterns for paths that allow editing files and folders. Defaults to `['**']`.
```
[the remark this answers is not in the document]
AriPerkkio: Default `**` is correct. Editing is disabled by default on `@tutorialkit/astro` level. On `@tutorialkit/react` level, it is enabled when you pass `onFileChange` prop. 

Maybe this could be

```
* `allowEditPatterns?: string[]` - Glob patterns for paths that are allowed to call `onFileChange`. Defaults to `['**']`.
```

packages/react/src/Panels/WorkspacePanel.tsx
AriPerkkio: Yep, it was implemented like this in https://github.com/stackblitz/tutorialkit/pull/314 as we didn't have `allowEditPatterns`. It was simply on/off. Note that `allowEditPatterns` is internal property at this level. TutorialKit authors use `editor.fileTree.allowEdits` only.

So do you mean that in `@tutorialkit/react`, when users use `<FileTree>` directly, we should disable calling given `props.onFileChange` when users don't explicitly pass `props.allowEditPatterns`?

Currently it works like this:

```jsx
// files are not editable
<FileTree />

// any file is editable
<FileTree onFileChange={console.log} />

// specific files are editable
<FileTree onFileChange={console.log} allowEditPatterns={["**.ts"]} />
```

But with that proposed change it could be

```jsx
// files are not editable
<FileTree />

// files are not editable, onFileChange is not called
<FileTree onFileChange={console.log} />

// specific files are editable
<FileTree onFileChange={console.log} allowEditPatterns={["**.ts"]} />
```

Nemikolh: > So do you mean that in `@tutorialkit/react`, when users use `<FileTree>` directly, we should disable calling given `props.onFileChange` when users don't explicitly pass `props.allowEditPatterns`?

Correct!

In your example, I find the second version to be the least suprising because it matches the same experience that you get with a `<button>` for instance. A button would not be in a disabled state when it doesn't have any callback provided and suddenly become enabled when it has one.

Here it's the same idea (albeit a negated version). Whether `onFileChange` is undefined or not should not impact whether edits are allowed or not. This should be driven entirely by the `allowEditPatterns` property IMO.

packages/react/src/core/ContextMenu.tsx
```diff
@@ -184,21 +168,19 @@ function MenuItem({ icon, children, ...props }: { icon: string } & ComponentProp
...
-            <RadixDialog.Title className="text-6 mb-2">Error</RadixDialog.Title>
+            <RadixDialog.Title className="text-6">{title}</RadixDialog.Title>
 
-            {children}
+            <div className="my-4">{children}</div>
 
-            <RadixDialog.Close title="Close" className="absolute top-4 right-4 w-6 h-6">
-              <span aria-hidden className="i-ph-x block w-full h-full"></span>
-            </RadixDialog.Close>
+            <RadixDialog.Close className="px-3 py-1 border border-tk-border-primary rounded">OK</RadixDialog.Close>
```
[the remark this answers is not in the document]
AriPerkkio: Done, now it looks like this

packages/astro/src/default/styles/variables.css
Nemikolh, line 1: Nice, huge improvement!! :heart_eyes:

packages/astro/src/default/components/LoginButton.tsx
Nemikolh, line 1: Gosh you've got no idea how much I love this change :tada:

packages/react/src/Button.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export const Button = forwardRef(({ className, variant = 'primary', ...props }: Props, ref: Ref<HTMLButtonElement>) => {
+  return (
+    <button
+      ref={ref}
+      {...props}
+      className={classNames(
+        className,
+        'flex items-center font-500 text-sm ml-2 px-4 py-1 rounded-md disabled:opacity-32',
```
Nemikolh, line 15:
```suggestion
        'flex items-center font-500 text-sm px-4 py-1 rounded-md disabled:opacity-32',
```

packages/react/src/core/ContextMenu.tsx
```diff
@@ -54,7 +56,7 @@ export function ContextMenu({
   const [state, setState] = useState<'idle' | 'add_file' | 'add_folder' | 'add_failed'>('idle');
   const inputRef = useRef<HTMLInputElement>(null);
 
-  if (!onFileChange) {
+  if (!onFileChange || !allowEditPatterns?.length) {
```
Nemikolh, line 59:
```suggestion
  if (!allowEditPatterns?.length) {
```

(so that the context menu still shows up which is useful when playing around with the component)

packages/react/src/core/ContextMenu.tsx
```diff
@@ -180,7 +182,9 @@ function Dialog({ title, onClose, children }: { title: string; onClose: () => vo
 
             <div className="my-4">{children}</div>
 
-            <RadixDialog.Close className="px-3 py-1 border border-tk-border-primary rounded">OK</RadixDialog.Close>
+            <RadixDialog.Close asChild>
+              <Button>OK</Button>
```
Nemikolh, line 186: How about we make it a bit larger? Wdyt?

```suggestion
              <Button className="min-w-20 justify-center">OK</Button>
```

Oh also we probably need `OK` to be `i18n?.confirmationText || DEFAULT_LOCALIZATION.confirmationText`

packages/react/src/utils/classnames.ts
```diff
@@ -5,7 +5,7 @@
  * @link http://jedwatson.github.io/classnames
  */
 
-type ClassNamesArg = undefined | string | Record<string, boolean> | ClassNamesArg[];
+type ClassNamesArg = undefined | false | string | Record<string, boolean> | ClassNamesArg[];
```
Nemikolh, line 8: neat :pinched_fingers:

packages/astro/src/default/components/LoginButton.tsx
```diff
@@ -48,21 +48,8 @@ export function LoginButton() {
...
-          !showLogin,
-        'hover:bg-tk-elements-topBar-primaryButton-backgroundColorHover hover:text-tk-elements-topBar-primaryButton-textColorHover':
-          !disabled && showLogin,
-        'hover:bg-tk-elements-topBar-secondaryButton-backgroundColorHover hover:text-tk-elements-topBar-secondaryButton-textColorHover':
-          !disabled && !showLogin,
-      })}
-      disabled={disabled}
-      onClick={onClick}
-    >
+    <Button variant={showLogin ? 'primary' : 'secondary'} disabled={disabled} onClick={onClick}>
```
Nemikolh, line 51:
```suggestion
    <Button className="ml-2" variant={showLogin ? 'primary' : 'secondary'} disabled={disabled} onClick={onClick}>
```

Nemikolh (approved):
Yay! 🤩🤩

PR #333, opened 2024-09-16
chore(dev-deps): update `eslint`, `prettier`, `vitest`

AriPerkkio:
- chore(dev-deps): update `@blitz/eslint-plugin`
- chore(dev-deps): update `eslint-plugin-astro`
- chore(dev-deps): update prettier packages
- chore(dev-deps): update `vitest`
- chore(dev-deps): remove `vite-plugin-inspect`


packages/astro/package.json
```diff
@@ -69,8 +69,7 @@
     "esbuild-node-externals": "^1.13.1",
     "execa": "^9.2.0",
     "typescript": "^5.4.5",
-    "vite-plugin-inspect": "0.8.4",
```
Nemikolh, line 72: This plugin has been pretty handy when debugging issues with virtual modules, it's probably fine if we keep it around?

AriPerkkio: I would rather install it only when needed but sure I can revert this if we want to keep it around.

AriPerkkio: Yep, the reason why I wanted to remove is this that you cannot use locally linked packages while running dev mode on TutorialKit monorepo. Linked projects will run into errors like `Cannot find package 'vite-plugin-inspect' imported from /x/tutorial-vite-plugin/node_modules/.pnpm/file+..+tutorialkit+packages+astro_@types+react-dom@18.3.0_astro@4.15.6_postcss@8.4.47_vite@5.4.5/node_modules/@tutorialkit/astro/dist/index.js`. But the solution for that is to not use dev/watch mode, and run manually `pnpm build` after making changes.

AriPerkkio: 1. Setup another Tutorial project with https://github.com/stackblitz/tutorialkit/blob/main/CONTRIBUTING.md#testing-tutorialkit-against-external-packages

2.1. Build linked monorepo with `pnpm dev`
2.2. Build or start dev mode of reference project
2.3. -> Error, no `vite-plugin-inspect`

3.1. Build linked monorepo with `pnpm build`
3.2. Build or start dev mode of reference project
3.3. -> Works fine

Nemikolh: Ah I see! Yeah that's annoying. Have you try doing:

```bash
rm -rf packages/template/src/content
ln -s ../path/to/other/tutorial/src/content ./packages/template/src/content
pnpm dev
```

If that works, maybe we could rename the `template/src/content` to something else and generate a symlink at build time (we could use `mklink` on Windows).

Alternatively, we could have this plugin behind a different environment variable that is not set by default (but could be set with a `.env` or equivalent).

AriPerkkio: Linking just the `content` directories makes Astro throw `UnknownContentCollectionError` there.

Also with projects like Vite Plugin Tutorial it's not just as simple as linking the `content`. We would need templates, `astro.config.ts`, `theme.css`, override components, etc - pretty much whole project.

Nemikolh: Ok, let's put this import behind a different environment variable that has to be set manually then :+1:

AriPerkkio: Added `VITE_INSPECT` for this.

Nemikolh:
Nice! You also need to update the `build.js` script:

https://github.com/stackblitz/tutorialkit/blob/afbc81f241ea471eaf334514ad4ac2fa96bc3662/packages/astro/scripts/build.js#L47-L49

This should be:

```ts
    define: {
      'process.env.TUTORIALKIT_DEV': JSON.stringify(process.env.TUTORIALKIT_DEV ?? null),
      'process.env.VITE_INSPECT': JSON.stringify(process.env.VITE_INSPECT ?? null),
    },
```

Also wondering if we should call it `TUTORIALKIT_VITE_INSPECT`. Thoughts?

AriPerkkio:
> Nice! You also need to update the `build.js` script:

Interesting way of using Vite's `define` - it overrides `process.env.<var>` from environment variables. I guess this is so that you cannot use these environment variables after build? 

Another interesting note: this is what dead-code-elimination does to this line. I guess there's some edge case why it has to leave `await` there too:

```diff
- process.env.TUTORIALKIT_DEV ? (await import('vite-plugin-inspect')).default() : null,
+ null ? (await null).default() : null
```

https://unpkg.com/browse/@tutorialkit/astro@0.2.3/dist/index.js

Nemikolh:
Oh that's an odd one! 

I've been trying to get esbuild to generate the same thing in their playground but failed. :open_mouth: 

[See this esbuild playground link.](https://esbuild.github.io/try/#YgAwLjE4LjEyAHsKICBidW5kbGU6IHRydWUsCiAgZXh0ZXJuYWw6IFsndGVzdCddLAogIGRlZmluZTogeyB0ZXN0OiAnbnVsbCcgfSwKfQBlAGVudHJ5LmpzAGNvbnN0IGJheiA9IHRlc3Q7Cgphc3luYyBmdW5jdGlvbiB0ZXN0KCkgewogIGNvbnN0IGZvbyA9IGJheiA/IGF3YWl0IGltcG9ydCgndGVzdCcpIDogZmFsc2U7CgogIGlmIChiYXopIHsKICAgIGNvbnNvbGUubG9nKCdEQ0VkJyk7CiAgfQoKICBpZiAodGVzdCkgewogICAgY29uc29sZS5sb2coJ0RDRWQnKTsKICB9Cn0)

It always eliminate the `await`, even with `minify: false` which is what we do for Astro. :thinking:

Nemikolh:
> Interesting way of using Vite's `define` - it overrides `process.env.<var>` from environment variables. I guess this is so that you cannot use these environment variables after build?

Btw, this is not using `vite`, we use `esbuild` directly for our Astro integration. And yes, the idea is that those environment variable cannot be used after a build.

We do this because `vite-inspect-plugin` is a dev dependency of our Astro integration. So the the dynamic import would (always?) fail for our end users if they had those env vars defined by accident.

Nemikolh (approved):
Nice! Let's merge this :raised_hands:

PR #334, opened 2024-09-16
feat: sync files from WebContainer to editor

### LINKED ISSUE #208: Update _files content with scrips command

Nemikolh:
This PR adds a new property `filesystem` to the metadata of a Tutorial / Chapter / Lesson:

```yaml
filesystem:
  syncChanges: true # default to false
```

When `syncChanges` is set to `true`, if files are modified on WebContainer via a terminal or by a process spawned manually via the `webcontainer` export on `tutorialkit:core`, the changes will be reflected in the editor.

Closes #208

This is opt-in because it can have a performance impact on the UI if lots of files are being modified / generated.

In the future we will likely accept `boolean | string` for `syncChanges` to be able to narrow down the watched files and improve performance.

### DIFF
--- a/packages/runtime/src/store/tutorial-runner.ts
+++ b/packages/runtime/src/store/tutorial-runner.ts
@@ -1,11 +1,13 @@
 import type { CommandsSchema, Files } from '@tutorialkit/types';
-import type { WebContainer, WebContainerProcess } from '@webcontainer/api';
-import type { TerminalStore } from './store/terminal.js';
-import { newTask, type Task, type TaskCancelled } from './tasks.js';
-import { clearTerminal, escapeCodes, type ITerminal } from './utils/terminal.js';
-import { Command, Commands } from './webcontainer/command.js';
-import { StepsController } from './webcontainer/steps.js';
-import { diffFiles, toFileTree } from './webcontainer/utils/files.js';
+import type { WebContainer, WebContainerProcess, IFSWatcher } from '@webcontainer/api';
+import type { TerminalStore } from './terminal.js';
+import { newTask, type Task, type TaskCancelled } from '../tasks.js';
+import { clearTerminal, escapeCodes, type ITerminal } from '../utils/terminal.js';
+import { Command, Commands } from '../webcontainer/command.js';
+import { StepsController } from '../webcontainer/steps.js';
+import { diffFiles, toFileTree } from '../webcontainer/utils/files.js';
+import type { EditorStore } from './editor.js';
+import { MultiCounter } from '../utils/multi-counter.js';
 
 interface LoadFilesOptions {
   /**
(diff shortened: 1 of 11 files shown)


Nemikolh:
Thanks for the questions! :smiley: Answering inline:

> Does this enable users to create files from terminal/commands so that they appear in file editor?

It does not at the moment. I was thinking we would do this later. Maybe as part of the `filesystem.visible` work you started?

What do you think?

> If not, maybe this doesn't close #208?

I think it does. Or at least based on what the first comment says:

> I'm creating a NgRx (Angular state management library) and the first step is to add the library with the `ng add @ngrx/store@latest`. It adds the dependency in package.json and updates some files.
>
> I added all these files in the _files folder as the student task is to run the script and sees how it affect the project.
> However even if the command succeeds, the files are not updated.

I think we can consider it closed when we merge this PR.

> We should keep https://github.com/stackblitz/tutorialkit/pull/165 in mind during this PR. I think that could be `filesystem.visible: string | string[]`.

Oh I really like that naming! :star2: 

> Naming of `filesystem.syncChanges`. As this could in future also allow globs for **watched** files, should we rename this to `filesystem.watch` instead?

Ah yeah this makes more sense, I'll rename it :+1:

packages/runtime/src/store/tutorial-runner.ts
```diff
@@ -503,6 +552,77 @@ export class TutorialRunner {
...
+        }
+      });
+    };
+
+    const scheduleReadFor = (filePath: string, encoding: 'utf-8' | null) => {
+      filesToRead.set(filePath, encoding);
+
+      clearTimeout(timeoutId);
+      timeoutId = setTimeout(readFiles, 100);
+    };
```
AriPerkkio, line 600: Could we add comment here stating what this does, something like `"// add file to read-queue and discard previously queued read task`. 

I had to check manually that this doesn't accidentally cancel the previous reads if multiple ones are done within 100ms. 

Really clever and compact scheduler btw 💯

AriPerkkio:
> > Does this enable users to create files from terminal/commands so that they appear in file editor?
> 
> It does not at the moment. I was thinking we would do this later. Maybe as part of the `filesystem.visible` work you started?
> 
> What do you think?

Sounds good to me, let's leave it to the next PR.

packages/runtime/src/store/tutorial-runner.ts
Nemikolh: Definitely! Very good point! Thanks for checking it :smiley:

e2e/src/components/ButtonWriteToFile.tsx
```diff
@@ -1,13 +1,23 @@
 import tutorialStore from 'tutorialkit:store';
+import { webcontainer } from 'tutorialkit:core';
 
 interface Props {
   filePath: string;
   newContent: string;
+  useWebcontainer?: boolean;
```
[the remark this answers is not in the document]
Nemikolh: Very good point!

Nemikolh:
Oh very good point!!

Dang, I knew there was something missing with testing and this PR :sweat_smile: , I'll add it :+1:

AriPerkkio (approved):
Test case looks good, thanks for adding it! 💯

PR #336, opened 2024-09-19
fix: prevent overwriting template files via `<FileTree>`

AriPerkkio:
Show an error if user tries to create file that already exists on file system, but is not visible on FileTree. Such files are for example non-lesson template files.

### DIFF
--- a/packages/react/src/core/ContextMenu.tsx
+++ b/packages/react/src/core/ContextMenu.tsx
@@ -1,6 +1,6 @@
 import { Root, Portal, Content, Item, Trigger } from '@radix-ui/react-context-menu';
 import * as RadixDialog from '@radix-ui/react-dialog';
-import { DEFAULT_LOCALIZATION, type FileDescriptor, type I18n } from '@tutorialkit/types';
+import { DEFAULT_LOCALIZATION, type FileDescriptor, type I18n, type FilesystemError } from '@tutorialkit/types';
 import picomatch from 'picomatch/posix';
 import { useRef, useState, type ComponentProps, type ReactNode } from 'react';
 import { Button } from '../Button.js';
@@ -18,8 +18,8 @@ interface FileRenameEvent extends FileChangeEvent {
 }
 
 interface Props extends ComponentProps<'div'> {
-  /** Callback invoked when file is changed. */
-  onFileChange?: (event: FileChangeEvent | FileRenameEvent) => void;
+  /** Callback invoked when file is changed. This callback should throw errors with {@link FilesystemError} messages. */
+  onFileChange?: (event: FileChangeEvent | FileRenameEvent) => Promise<void>;
 
   /** Glob patterns for paths that allow editing files and folders. Disabled by default. */
   allowEditPatterns?: string[];
@@ -37,6 +37,7 @@ interface Props extends ComponentProps<'div'> {
     | 'fileTreeCreateFolderText'
     | 'fileTreeActionNotAllowedText'
     | 'fileTreeAllowedPatternsText'
+    | 'fileTreeFileExistsAlreadyText'
     | 'confirmationText'
   >;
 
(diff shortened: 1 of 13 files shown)


packages/runtime/src/store/tutorial-runner.ts
```diff
@@ -189,6 +189,62 @@ export class TutorialRunner {
...
+        async (signal) => {
+          await previousLoadPromise;
+
+          const webcontainer = await this._webcontainer;
+
+          if (signal.aborted) {
+            reject(new Error('Task was aborted'));
+          }
+
+          signal.throwIfAborted();
```
AriPerkkio, line 206: @Nemikolh, extra attention here please 😄 
Is this correct?

packages/types/src/entities/index.ts
```diff
@@ -6,6 +6,8 @@ export type * from './nav.js';
 export type FileDescriptor = { path: string; type: 'file' | 'folder' };
 export type Files = Record<string, string | Uint8Array>;
 
+export type FilesystemError = 'FILE_EXISTS' | 'FOLDER_EXISTS';
```
AriPerkkio, line 9: I thought about using real `Error` here, like `class FilesystemError extends Error {}` but maybe we can use something as simple as string literals here? Any thoughts?

packages/runtime/src/store/tutorial-runner.ts
Nemikolh: It is! :raised_hands: 

Thinking more about this, given what this method does, I think that we should just remove everything related to cancellation here. It's mostly there to speed up when moving between lessons quickly.

packages/runtime/src/store/tutorial-runner.ts
```diff
@@ -189,6 +189,62 @@ export class TutorialRunner {
...
+          } catch {
+            resolve(false);
+          }
+        },
+        { ignoreCancel: true },
+      );
+    });
+  }
+
+  async folderExists(folderPath: string) {
```
Nemikolh, line 220: Oh maybe we could create a helper function that both `folderExists` and `fileExists` call? Wdyt?

packages/types/src/entities/index.ts
Nemikolh: I think this is fine, we could also use an enum but a string literal is fine by me :+1:

packages/react/src/core/ContextMenu.tsx
```diff
@@ -72,14 +75,22 @@ export function ContextMenu({
...
+      try {
+        await onFileChange?.({
           value,
           type: state === 'add_file' ? 'file' : 'folder',
           method: 'add',
         });
-      } else {
-        return setState('add_failed');
+      } catch (error: any) {
+        const message: FilesystemError | (string & {}) | undefined = error?.message;
```
Nemikolh, line 89: Oh first time seeing this kind of types. What benefit does `FilesystemError | (string & {}) | undefined` has?

Nemikolh (commented):
Changes looks good!

Seeing the screenshot, I wonder if we should add more padding on the y-axis? What do you think?

packages/react/src/core/ContextMenu.tsx
AriPerkkio: It's for IDE auto completion: https://www.typescriptlang.org/play/?#code/C4TwDgpgBA0gdgewO5wCrggZygXigIgDMBLAJ02HygB8DMIBjBOAE3wG4AoT0SKAQQA2g5EMEBlYKWJwA5rliIU6SNloVpcmlACurCCTgQWXJnApREASTjAIw4vXMQAXAOGjhkzfLz4O3LzQYp6C-KwACqQIAG7ELBD8OsAIAMIIALZgghB2CvDIaBhqUBoy8gBkUADeAL7aegmGxqbMFgAWAIaYNnYOTvRuIUhi4SxRsfGJyWmZ2bnQfgGcAPQrUACysdCECAw62ABGuUgQEHAEVJ2sUFLEsrIQpFCdM1BMWTl5SMTA7VCpDYAEQA1JgwJ0GNB1qlUAAlAAyYIhUM4QA

Here we expect that any kind of errors can come up, but want to make sure we don't typo types of `FilesystemError`.

Nemikolh: Wooo this is super coool! :star_struck: TIL.

packages/runtime/src/store/tutorial-runner.ts
```diff
@@ -189,6 +189,62 @@ export class TutorialRunner {
     );
   }
 
+  async fileExists(filepath: string) {
+    const previousLoadPromise = this._currentLoadTask?.promise;
```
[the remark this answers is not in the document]
AriPerkkio: Good point! I'll add that. 👍

AriPerkkio:
> Seeing the screenshot, I wonder if we should add more padding on the y-axis?

Sure! Now it looks like this:

Nemikolh (approved):
Good job on this! :tada:

PR #342, opened 2024-09-20
fix(astro): better default meta tags

eric-burel:
This PR proposes to use the site logo as the default open graph image, and sets up better default meta tags for opengraph and twitter.

(please double check as opengraph display is usually tested on deployed instances, I couldn't test them locally aside from reading them in the browsers element tab)

Btw I don't think the BASE_URL env variable is documented, to have a proper absolute URL for meta tags.


AriPerkkio:
I'll take a look at the code next week, but another idea came into my mind: Should we create new `MetaData` component that authors could use to define completely custom meta tags? https://tutorialkit.dev/guides/overriding-components/

So something like:

```ts

export default defineConfig({
  integrations: [
    tutorialkit({
      components: {
        MetaData: './src/components/MetaTags.astro',
      },
    }),
  ],
});
```

And then in `./src/components/MetaTags.astro` you would get access to default meta tags, and you could decide whether to use them or not. And of course add your custom ones there too.

Improving default meta tags is important too.

eric-burel:
Definitely, I've thought about a Metadata component but then the question is how to pass the settings from a lesson to this component
Maybe using the lesson slug as a unique id
This also joins the question of accessing lesson metadata from the content collection to generate tables of content rss feeds etc.

AriPerkkio (approved):
Looks good to me, but as I'm not super familiar with Open Graph, @Nemikolh & @sulco what do you think? 

I think @sulco you mentioned something about social preview picture generation in the past. 🤔

packages/astro/src/default/layouts/Layout.astro
```diff
@@ -4,19 +4,27 @@ import { joinPaths } from '../utils/url';
 
 interface Props {
   title: string;
+  description?: string;
 }
 
-const { title } = Astro.props;
+const { title, description = 'TutorialKit' } = Astro.props;
 const baseURL = import.meta.env.BASE_URL;
+const logoUrl = joinPaths(baseURL, '/logo.svg');
```
Nemikolh, line 12: We should probably move the logic from `Logo.astro` in an utility function as we support multiple formats. Here the logo has to be ending with `.svg` which might not be the case.

So you want to copy the logic from here:

https://github.com/stackblitz/tutorialkit/blob/c8ce1826aeb79ef878a4e5e7104369ce900eb76d/packages/astro/src/default/components/Logo.astro#L12-L26

To a new file, like `src/default/utils/logo.ts`. You'll want to move the `LOGO_EXTENSIONS` constant as well.

packages/astro/src/default/pages/[...slug].astro
```diff
@@ -17,7 +17,7 @@ type Props = InferGetStaticPropsType<typeof getStaticPaths>;
 const { lesson, logoLink, navList, title } = Astro.props as Props;
 ---
 
-<Layout title={title}>
+<Layout title={title} description="A TutorialKit interactive lesson">
```
Comment (Nemikolh) [inline packages/astro/src/default/pages/[...slug].astro:20]: Ideally this should come from the metadata. You can grab it from `Astro.props` here and then it needs to be returned as part of `generateStaticRoutes`.

Nemikolh (commented):
I like this idea a lot! Thanks for the PR! :smiley: 

I've left a couple of things to address. @AriPerkkio Or do you want to merge this as is and address the feedback in a follow-up PR?

packages/astro/src/default/pages/[...slug].astro
Comment (eric-burel) [reply packages/astro/src/default/pages/[...slug].astro:20]: @Nemikolh I would advise to improve the collection, rather than `generateStaticRoutes`, to allow feeding more data to the "lesson" object. 
Because `generateStaticRoutes` is specific to this page only and is difficult to reuse if you want to craft a sitemap or an RSS feed. And you still need a way to pass some metadata from your MDX file to feed "generateStaticRoutes". That's why I couldn't come up with a more customizable/elegant solution yet.

packages/astro/src/default/layouts/Layout.astro
eric-burel: Yep done, I've also crafted an utility for the favicon (with only SVG support). And documented BASE_URL in the deployment section of the doc, to explain how to get absolute URLs that are needed for open graph.

packages/astro/src/default/pages/[...slug].astro
Comment (Nemikolh) [reply packages/astro/src/default/pages/[...slug].astro:20]: Hey! I think I should have provided more pointers, sorry!

Let me explain. So here's what we need to do:

* Add a `metaTagsSchema` to https://github.com/stackblitz/tutorialkit/blob/main/packages/types/src/schemas/common.ts:

```ts
export const metaTagsSchema = z.object({
  // ideally we would have want to use image from 
  // https://docs.astro.build/en/guides/images/#images-in-content-collections
  // but we can't so we'll need a custom logic (see below)
  image: z.string().optional().describe('An image to show on social previews'),
  description: z.string().optional().describe('...'),
  title: z.string().optional().describe('...'),
})

...

export const webcontainerSchema = commandsSchema.extend({
  meta: metaTagsSchema.optional()
    .describe('Configure Open graph meta tags for previews on social medias'),
  ...
});
```

* Add `'meta'` here:

https://github.com/stackblitz/tutorialkit/blob/f49e9107d35b98079a0fb16c74b9f37a45357661/packages/astro/src/default/utils/content.ts#L238-L240

* Now we could handle `image` in the same file when the collection is traversed to resolve all the relative path to be fully resolved. However that won't be enough to have them included in the final bundle. Instead for now we can limit ourselves to images residing in the `public` folder (so starting with `/`).

* Finally update `[...slug].astro` to use the new `lesson.meta?.<prop>` properties

Comment (eric-burel) [reply packages/astro/src/default/pages/[...slug].astro]: @Nemikolh right thanks for the insight, I was not sure if this should be implemented in user-land or not but it seems a sound idea to have at least a title, description and image default setting, having a finer grained system (eg you want a different image for OpenGraph and Twitter, or images outsides of /public or even generated via an API) would be worth a different PR I think

eric-burel:
I have a small doubt on the update I've done in the documentation : I documented the BASE_URL variable, which is used to compute the logo, favicon or public images. Initially I thought this would the absolute URL of the application (you want absolute urls for open graph tags) but I am not sure. It seems to be a variable set by Astro and which might rather be a kind of separate directory, eg to handle a monorepo.
But I am not sure, wha'ts your opinion on this @Nemikolh ?

packages/astro/src/default/pages/[...slug].astro
Comment (Nemikolh) [reply packages/astro/src/default/pages/[...slug].astro]: I agree! Yeah, let's start with a simple set of settings and move to a finer grained system later :+1:

eric-burel:
I still don't exactly get what `BASE_URL` is, but I figured that `Astro.site` is what I want for absolute URLs, I've pushed an update in this direction.
It's not super elegant but the SEO tags are at least correct.

Nemikolh:
Hey @eric-burel !

Do you think you could rebase this PR on top of the latest version of main? This would help reviewing this.

So you do `git fetch <stackblitz-tk-remote>` and then `git rebase <stackblitz-tk-remote>/main`. Trying this, you should only get a single conflict on `packages/astro/src/default/layouts/Layout.astro`.

If you want I can also do it for you, but that means that I'll force push to your branch and you'll have to reset it to the origin. Let me know what you prefer :raised_hands:

eric-burel:
@Nemikolh done, I think I've messed up at some point and used a "merge" instead. It has fixed the diff :ok_hand:

eric-burel:
I've also updated the logic to access the site absolute URL, I found another doc explaining that `Astro.site` will feed the `import.meta.env.SITE` env variable for consumption in JS files https://docs.astro.build/en/guides/environment-variables/
Ideally this would be set to "http://localhost:4321" in dev but I am not sure how to reliably get the right port. 
Astro.url automatically has an absolute URL in development but it's not the right solution, because it's based on the current HTTP request to get a full path while we just want the website root URL so the site value.

Nemikolh:
Looking at the PR again, I would keep the previous usage of `BASE_URL` in place of `site` for resources loaded by the browser (so everything that is not OG or Twitter related) as those links can be relative and don't have to be absolute.

A big downside of forcing them to be absolute is that it makes preview deployment with cloudflare or netlify a bit more annoying as they would need a custom build.

Regarding your question on what `BASE_URL` is, see the docs on [`base`](https://docs.astro.build/en/guides/deploy/github/#base).

packages/astro/src/default/components/MetaTags.astro
```diff
@@ -0,0 +1,36 @@
...
+imageUrl ??= readLogoFile('logo', true);
+const faviconUrl = readFaviconFile('favicon', true);
+---
+
+<meta charset="UTF-8" />
+<meta name="viewport" content="width=device-width" />
+<meta name="generator" content={Astro.generator} />
+{meta.description ? <meta name="description" content={meta.description} /> : null}
+{meta.title ? <title>{meta.title}</title> : null}
+{faviconUrl ? <link rel="icon" type="image/svg+xml" href={faviconUrl} /> : null}
```
Nemikolh, line 28: This should keep using the `import.meta.BASE_URL`. Also this isn't a meta tag, like `<title>`, let's move them back to where they were.

docs/tutorialkit.dev/src/content/docs/guides/deployment.mdx
```diff
@@ -34,6 +34,17 @@ This will generate a `dist` directory containing the static files that make up y
 
 You can learn more about the build process in the [Astro documentation](https://docs.astro.build/en/reference/cli-reference/#astro-build).
 
+## Environment variables
+
+The [`site`](https://docs.astro.build/fr/reference/configuration-reference/#site) configuration should point to your website's absolute URL.
+This will allow to compute absolute URLs for SEO metadata.
+
+Example:
+```json
```
Nemikolh, line 43:
```suggestion
```js
```

packages/astro/src/default/components/MetaTags.astro
```diff
@@ -0,0 +1,36 @@
...
+}
+imageUrl ??= readLogoFile('logo', true);
+const faviconUrl = readFaviconFile('favicon', true);
+---
+
+<meta charset="UTF-8" />
+<meta name="viewport" content="width=device-width" />
+<meta name="generator" content={Astro.generator} />
+{meta.description ? <meta name="description" content={meta.description} /> : null}
+{meta.title ? <title>{meta.title}</title> : null}
```
Nemikolh, line 27: Should `title` and `og:title` always be the same? I'm tempted to think we should keep them separate.

@AriPerkkio what do you think?

packages/astro/src/default/utils/routes.ts
```diff
@@ -26,6 +26,7 @@ export async function generateStaticRoutes() {
             logoLink: tutorial.logoLink,
             navList: generateNavigationList(tutorial, import.meta.env.BASE_URL),
             title: `${part.data.title} / ${chapter.data.title} / ${lesson.data.title}`,
+            meta: lesson.data.meta,
```
Nemikolh, line 29: I would remove this as there's not much benefit compared to `lesson.data.meta`.

packages/astro/src/default/utils/favicon.ts
Nemikolh, line 1: I don't think this function buys us much. I would just use `readPublicImage` directly with `favicon.svg` as an argument.

packages/astro/src/default/layouts/Layout.astro
```diff
@@ -1,24 +1,18 @@
 ---
 import { ViewTransitions } from 'astro:transitions';
-import { joinPaths } from '../utils/url';
+import type { MetaTagsConfig } from '@tutorialkit/types';
+import MetaTags from '../components/MetaTags.astro';
 
 interface Props {
-  title: string;
```
Nemikolh, line 6: Let's keep `title` and add `meta` as a separate props.

Nemikolh (commented):
Looks pretty good! Good job on this :star2:

packages/astro/src/default/components/MetaTags.astro
eric-burel: Yeah handling the title raises some questions, title is used in the navbar and is currently computed as a breadcrumbs, while meta.title is what you want to ve visible on social media

AriPerkkio: Setting `docment.title` from `meta.title` doesn't sound right to me. So maybe it's better to have separate ones.

eric-burel:
Hi, I'll get back at the few last comments, but in the meantime I've opened a proposal for a portal system in Astro: https://github.com/withastro/roadmap/discussions/1032

This would make it easier to support any type of tag that lands into `<head>` without having to support a full-fledged intermediate JSON représentation that fits into the YAML frontmatter. Not that it's impossible, it's Next.js approach with the metadata object, but maybe more work than it should at framework-level for something that could be done in user-land with good old HTML.

I am thinking of alternate links for instance that I will need in AstroPatterns/NextPatterns because I translate similar content for 2 frameworks and may be punished by Google for that if I don't provide alternate links.

packages/astro/src/default/components/MetaTags.astro
eric-burel: Makes sense, I'll stick to the lesson's title then for the document title

packages/astro/src/default/components/MetaTags.astro
eric-burel: Yes it does use BASE_URL but wrapped within readPublicImage, so the path management is centralized in a single function

eric-burel:
> Looking at the PR again, I would keep the previous usage of `BASE_URL` in place of `site` for resources loaded by the browser (so everything that is not OG or Twitter related) as those links can be relative and don't have to be absolute.

`BASE_URL` is still there, and optionally the absolute option can be used to also append the whole website URL via `Astro.site`. `site` is only actually used when using the logo as the default ogImage, though I think having it as an option in `readPublicAsset` may be useful.

> A big downside of forcing them to be absolute is that it makes preview deployment with cloudflare or netlify a bit more annoying as they would need a custom build.

If you don't set `Astro.site` you'll just come up with a warning. I think you are expected to use an env variable in `astro.config` to handle these cases? Eg Vercel preview URLs.

Let me know if I got something wrong!

packages/astro/src/default/pages/[...slug].astro
```diff
@@ -15,9 +15,13 @@ export async function getStaticPaths() {
 type Props = InferGetStaticPropsType<typeof getStaticPaths>;
 
 const { lesson, logoLink, navList, title } = Astro.props as Props;
+const meta = lesson.data?.meta || {};
+// Use lesson's default title and a default description for SEO metadata
+meta.title ??= title;
+meta.description ??= 'A TutorialKit interactive lesson';
```
Comment (Nemikolh) [inline packages/astro/src/default/pages/[...slug].astro:21]: Nit:

```suggestion
const meta = lesson.data?.meta ?? {};

// use lesson's default title and a default description for SEO metadata
meta.title ??= title;
meta.description ??= 'A TutorialKit interactive lesson';
```

docs/tutorialkit.dev/src/content/docs/guides/deployment.mdx
```diff
@@ -34,6 +34,17 @@ This will generate a `dist` directory containing the static files that make up y
 
 You can learn more about the build process in the [Astro documentation](https://docs.astro.build/en/reference/cli-reference/#astro-build).
 
+## Environment variables
+
+The [`site`](https://docs.astro.build/fr/reference/configuration-reference/#site) configuration should point to your website's absolute URL.
```
AriPerkkio, line 39:
```suggestion
The [`site`](https://docs.astro.build/reference/configuration-reference/#site) configuration should point to your website's absolute URL.
```

packages/types/src/schemas/metatags.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const metaTagsSchema = z.object({
+  image: z
+    .string()
+    .optional()
+    /**
+     * Ideally we would have want to use image from:
+     * https://docs.astro.build/en/guides/images/#images-in-content-collections .
+     */
+    .describe('An image to show on social previews, should point to the public folder: "/foobar.png"'),
```
Nemikolh, line 11:
```suggestion
    /**
     * Ideally we would want to use `image` from:
     * https://docs.astro.build/en/guides/images/#images-in-content-collections .
     */
    .describe('A relative path to an image that lives in the public folder to show on social previews.'),
```

Nemikolh (approved):
Looks really good! Good work on this @eric-burel !

Just a few minor nits and it's good to go :raised_hands:

docs/tutorialkit.dev/src/content/docs/guides/deployment.mdx
eric-burel: You caught the French in me lol

packages/astro/src/default/utils/content.ts
```diff
@@ -245,6 +245,7 @@ export async function getTutorial(): Promise<Tutorial> {
           'editor',
           'focus',
           'i18n',
+          'meta',
```
AriPerkkio, line 248: `interface Lesson` has no `meta`

AriPerkkio (approved):
Looks good to me, thanks @eric-burel!

PR #348, opened 2024-09-25
docs: add example of `tutorialkit:store` and `tutorialkit:core` usages

AriPerkkio:
Adds documentation for `tutorialkit:store` and `tutorialkit:core`. 

- https://tutorialkit.dev/reference/tutorialkit-api
- https://tutorialkit.dev/guides/how-to-use-tutorialkit-api/


docs/tutorialkit.dev/astro.config.ts
```diff
@@ -67,6 +67,10 @@ export default defineConfig({
               label: 'Overriding Components',
               link: '/guides/overriding-components/',
             },
+            {
+              label: 'Using low level APIs',
+              link: '/guides/using-low-level-apis/',
+            },
```
AriPerkkio, line 73: This page will describe how to use `tutorialkit:store` and `tutorialkit:core`. What would be a good name for this page? Maybe something like:

- TutorialKit Node APIs
- Store and Code APIs

Feel free to suggest better name. 🙌

docs/tutorialkit.dev/src/content/docs/guides/using-low-level-apis.mdx
```diff
@@ -0,0 +1,189 @@
...
+
+export function TaskList() {
+  const [hasDefaultExport, setDefaultExport] = useState<boolean>(false);
+  const [hasSumFunction, setSumFunction] = useState<boolean>(false);
+
+  useEffect(() => {
+    const unsubscribe = subscribeFileChanges("/index.js", (content) => {
+      // Naive text based checking just for example. Use proper AST analysis in real lessons.
+      setDefaultExport(content.includes("export default"));
+      setSumFunction(content.includes("function sum"));
```
Nemikolh, line 150: This is a pretty cool idea but I'm a bit torn on this one.

On one hand, I'm not sure if it makes sense for us to recommend doing any form of pattern matching against an AST (especially in the UI thread).

But the other solution I have in mind, which is to run a program in webcontainer that import the file and run a test against it, and check the exit code of that process, is not going to be without its own issues too.

I suppose we should probably have example of both.

docs/tutorialkit.dev/src/content/docs/guides/using-low-level-apis.mdx
```diff
@@ -0,0 +1,189 @@
...
+  "node_modules",
+  "package-lock.json",
+  "package.json",
+  "vite.config.ts"
+]
+```
+
+### Example usage in React component
+
+Here is an example of custom React component that looks for updates in `/index.js` and provides user information about their lesson status.
```
Nemikolh, line 134: I think I would have this in its own "How to guide" to make it more discoverable. And maybe use a `babel` or `acorn` to be more useful.

docs/tutorialkit.dev/astro.config.ts
Nemikolh: So I would actually have more than one page:
* One under `reference` which could be called `TutorialKit API` which describes what is exposed by each module and maybe showing the functions similarly to https://webcontainers.io/api.
* One (but ideally more) under guides, that are _literal_ ["How to"](https://diataxis.fr/how-to-guides/)  guides explaining how to solve a specific problems that people have been running into.

So examples of what I think we should have (in no particular order):
- How to write custom components that access the tutorial state? _(basically what you've written so far)_
- How to programmatically write to one of the terminal? _(example component doing it via a button)_
- How to check the lesson was solved and congratz the user? _(your task example + https://github.com/stackblitz/tutorialkit/discussions/205)_
- How to watch a file? _(using `tutorialStore.onDocumentChanged`)_

docs/tutorialkit.dev/src/content/docs/guides/using-low-level-apis.mdx
```diff
@@ -0,0 +1,189 @@
...
+```ts
+import tutorialStore from "tutorialkit:store";
+```
+
+The Tutorial Store provides access to current state of the Tutorial, for example the files and their contents.
+
+```ts
+import tutorialStore from "tutorialkit:store";
+
+console.log(tutorialStore.documents.value);
```
Nemikolh, line 23: I think you're trying to be minimal here, but this is not code they should write.

I think we should always suggest using React components and have example that always wait for the lesson to be loaded via `useStore(tutorialStore.lessonFullyLoaded)`

(unless they run after a user action in which case they can ignore the wait).

docs/tutorialkit.dev/src/content/docs/guides/using-low-level-apis.mdx
```diff
@@ -0,0 +1,189 @@
...
+      <li>Add <code>function sum</code> to index.js {hasSumFunction ? "✅" : "❌"}</li>
+    </ol>
+  );
+}
+
+function subscribeFileChanges(filename: string, callback: (content: string) => void) {
+  const controller = new AbortController();
+
+  core.webcontainer.then((webcontainer) => {
+    const watcher = webcontainer.fs.watch(filename, async () => {
```
Nemikolh, line 168: We should recommend using `tutorialStore.onDocumentChanged` instead of recommending hooking into webcontainer. Doing this should always be last resort for users that know what they're doing.

First because it's going to be more efficient as it doesn't go through webcontainer.

But second, also because if the file is removed and then re-added, this would get stuck (but it should work with `onDocumentChanged` and if it doesn't then it's a bug).

Nemikolh (commented):
Thanks for starting this! I had a lot of thoughts on the topic so I wrote a LOT.

Sorry please bear with me :sweat_smile:

docs/tutorialkit.dev/src/content/docs/guides/using-low-level-apis.mdx
AriPerkkio: This was the only example I could think of from the `tutorialkit:core` module 😅 

But sure, happy to change this example with something else. As long as it's using `tutorialkit:core` which is the point here.

docs/tutorialkit.dev/astro.config.ts
AriPerkkio: Awesome, thanks! This makes the documentation task much clearer 😄 

So we will have these pages:

- Reference
  - TutorialKit API
    - Tutorial Store: describes all public APIs. We might want to mark all other ones with `@internal` JSDoc and use `stripInternal` in `tsconfig.json`.
    - Tutorial Core: describes how to access `webcontainer` and has link to webcontainers.io

- Guide
  - How to pages:
    - `Access tutorial state`. Uses `tutorialkit:store`. Do we show how to access file contents here like in the `HelpButton.tsx` or `TaskList.tsx`? 
    - `Write to terminal`. How is this done? `webcontainer.spawn`?
    - `Congrats the user when lesson is solved`. How to do this? There's no `lessonSolved`.
    - `How to watch a file`. `tutorialStore.onDocumentChanged`

docs/tutorialkit.dev/src/content/docs/guides/using-low-level-apis.mdx
Nemikolh: Yeah you're right I forgot for a minute that  those files should be pretty small 👍

docs/tutorialkit.dev/src/content/docs/guides/using-low-level-apis.mdx
Nemikolh: Oh sorry I completely missed the goal of that example.

For an example using `tutorialkit:core`, how about we have one spawning a process that "checks" whether or not the lesson was solved (by running a test) and if the process exit with zero we show a confettis?

docs/tutorialkit.dev/astro.config.ts
AriPerkkio: @Nemikolh I've just started to write the reference part. So something like this? 

<details>

```mdx
---
title: TutorialKit API
description: Use TutorialKit's lower level APIs
---

TutorialKit exposes low level APIs that authors can utilize to provide highly custom experience in their tutorials.

## Tutorial Store

You can access Tutorial Store by importing the `tutorialkit:store` entrypoint.

```ts
import tutorialStore from "tutorialkit:store";
```

### Common types

- `ReadableAtom` from [`nanostores`](https://www.npmjs.com/package/nanostores)

### Properties

#### `previews`

Instances of the preview tabs.

Type: `ReadableAtom<PreviewInfo[]>`

```ts
import type { PreviewSchema } from '@tutorialkit/types';

class PreviewInfo {
    readonly portInfo: PortInfo;
    title?: string;
    pathname?: string;
    get url(): string | undefined;
    get port(): number;
    get baseUrl(): string | undefined;
    get ready(): boolean;
    static parse(preview: Exclude<PreviewSchema, boolean>[0]): Preview;
    static equals(a: PreviewInfo, b: PreviewInfo): boolean;
}

interface Preview {
    port: number;
    pathname?: string;
    title?: string;
}
```

#### `terminalConfig`

Type: `ReadableAtom<TerminalConfig>`

#### `editorConfig`

Type: `ReadableAtom<EditorConfig>`

#### `currentDocument`

Type: `ReadableAtom<EditorDocument | undefined>`

#### `bootStatus`

Type: `ReadableAtom<BootStatus>`

#### `documents`

Type: `ReadableAtom<EditorDocuments>`

#### `files`

Type: `ReadableAtom<FileDescriptor[]>`

#### `template`

Type: `Files | undefined`

#### `selectedFile`

Type: `ReadableAtom<string | undefined>`

#### `lesson`

Type: `Readonly<Lesson> | undefined`

#### `themeRef`

Type: `ReadableAtom<unknown>`


### Methods


## Tutorial Core

```ts
import * as core from "tutorialkit:core";
```

...
```

</details>

Making all of this stuff part of public API is maybe too much. We'll need to always avoid making breaking changes in these APIs.

File: docs/tutorialkit.dev/src/content/docs/reference/tutorialkit-api.mdx [hunk 6]
```diff
@@ -0,0 +1,365 @@
...
+```
+
+:::note
+Using `tutorialkit:store` is **experimental** at the moment.
+This module may introduce breaking changes in patch and minor version updates. Pay extra attention when updating the versions.
+
+You can help us stabilize the API by providing feedback at [Stabilizing `tutorialkit:store` API | Github Discussions](https://github.com/stackblitz/tutorialkit/discussions).
+Please let us know how you are using this API.
+
+TODO UPDATE THIS LINK
```
AriPerkkio, line 24: TODO: Create discussion and update the link here.

packages/runtime/src/store/index.ts
```diff
@@ -351,10 +373,7 @@ export class TutorialStore {
     }
   }
 
-  updateFiles(files: Files) {
-    this._runner.updateFiles(files);
-  }
-
```
AriPerkkio, line 357: This and `refreshStyles` below are unused code. I though we might as well just remove them.

docs/tutorialkit.dev/src/content/docs/guides/using-low-level-apis.mdx
AriPerkkio: Now using `magicast` for this. https://stackblitz.com/edit/tutorialkit-api-use-helpbutton?file=src/components/HelpButton.tsx&file=src/content/tutorial/1-basics/1-introduction/1-welcome/content.mdx

Nemikolh: This is soooo cool! :heart_eyes: Good job on this! :smiley:

docs/tutorialkit.dev/src/content/docs/guides/how-to-use-tutorialkit-api.mdx
AriPerkkio, line 1: @Nemikolh do you think we need more examples? I left out the `How to watch a file? (using tutorialStore.onDocumentChanged)`, as `Provide feedback to user when lesson is solved` already uses that API. So the previous TaskList-example is no longer here.

AriPerkkio: Also cc. @sulco for brainstorming ⚡

Nemikolh: Woooow!! I think this is pretty amazing!! :heart_eyes: 

I love your examples here, they are really high quality :tada: 

I'd say we can merge this now as I find it already pretty useful for our users. We can always add more examples later. :smiley: 

Really good job on this @AriPerkkio :heart:

AriPerkkio: Awesome, thanks! Let's merge this and update the examples when needed. 🚀

PR #350, opened 2024-09-30
feat(runtime): add `terminal.input` for writing to stdin

AriPerkkio:
- During https://github.com/stackblitz/tutorialkit/pull/348 it was noticed that `terminal.write` cannot be used to write into terminal's stdin. It's only useful for writing to stdout.
- Adds new terminal method that can be used to write commands to stdin. The #348 will add proper example into documentation.

```tsx

export default function Example() {
  async function onClick() {
    const terminal = tutorialStore.terminalConfig.value!.panels[0];
    terminal.input('npm run test\n');
  }

  return (
    <button
      onClick={onClick}
      className="px-4 py-1 my-3 cursor-pointer border-1 border-tk-border-primary rounded-md bg-tk-elements-primaryButton-backgroundColor text-tk-elements-primaryButton-textColor"
    >
      Run tests
    </button>
  );
}
```

### DIFF
--- a/packages/runtime/src/webcontainer/terminal-config.ts
+++ b/packages/runtime/src/webcontainer/terminal-config.ts
@@ -67,6 +67,7 @@ export class TerminalPanel implements ITerminal {
   private _terminal?: ITerminal;
   private _process?: WebContainerProcess;
   private _data: string[] = [];
+  private _input: string[] = [];
   private _onData?: (data: string) => void;
 
   constructor(
@@ -127,9 +128,15 @@ export class TerminalPanel implements ITerminal {
       this._terminal.reset();
     } else {
       this._data = [];
+      this._input = [];
     }
   }
 
+  /**
+   * Write data to stdout.
+   *
+   * @internal
+   */
   write(data: string) {
     if (this._terminal) {
       this._terminal.write(data);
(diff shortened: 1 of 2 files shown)


packages/runtime/src/webcontainer/terminal-config.ts
```diff
@@ -127,9 +128,15 @@ export class TerminalPanel implements ITerminal {
       this._terminal.reset();
     } else {
       this._data = [];
+      this._input = [];
     }
   }
 
+  /**
+   * Write data to stdout.
```
AriPerkkio, line 136: Now that I'm looking at this second time - do we actually want to use `stdout` `stdin` concepts in terminal? These can also be output-only terminals which are not really terminals, just text outputs. Thoughts @Nemikolh?

Nemikolh: I think we shouldn't as this method does not write to `stdout` but only to the terminal.

packages/runtime/src/webcontainer/terminal-config.ts
```diff
@@ -67,6 +67,7 @@ export class TerminalPanel implements ITerminal {
   private _terminal?: ITerminal;
   private _process?: WebContainerProcess;
   private _data: string[] = [];
+  private _input: string[] = [];
```
Nemikolh, line 70: To be correct we should have a single array of `{ data: string, type: 'input' | 'echo' }`. That way we preserve the order of both (otherwise we will get them in out of order).

AriPerkkio: Good catch, I'll add that!

Nemikolh (approved):
Looks good! Glad we found and fixed that bug! :star_struck:

Issue #300, opened 2024-09-02
`Toggle Terminal` button should be hidden when `terminal: false`


AriPerkkio:
### Describe the bug

When terminal is disabled in metadata, the `Toggle Terminal` button is still visible. It just shows empty blank terminal when toggled.

### Steps to reproduce

1. Define in metadata:
  ```yaml
  ---
  type: lesson
  title: Test
  terminal: false
  ---
  ```
2. Open page and click `Toggle Terminal`

### Expected behavior

Whole `Toggle Terminal` button should be hidden. 

### Screenshots

![image](https://github.com/user-attachments/assets/143621a0-bd9b-46bf-86fc-db8d95f34482)

### Platform

- TutorialKit version: 0.2.1

Charan167:
Hello @AriPerkkio. I was trying to recreate this issue but it seems to be working fine. Could you please help me out?

https://github.com/user-attachments/assets/924cf1a8-c0f2-4161-a268-cb8b4e9336e1

AriPerkkio:
> but it seems to be working fine

Great, thanks for checking this! 🙌 

I'll check later if this issue is still valid. Feel free to proceed with the other issue mentioned in https://github.com/stackblitz/tutorialkit/pull/343#pullrequestreview-2322783870.

AriPerkkio:
> I'll check later if this issue is still valid. 

Yup, this is still valid. Added failing test case at https://github.com/stackblitz/tutorialkit/tree/test/terminal-false.

Task assigned to @RonithManikonda.

PR #363, opened 2024-10-03
Fix toggle terminal visibility

SeanHersh:
Description of What Happened:
In my open-source contribution, I aimed to fix a bug where the "Toggle Terminal" button stayed visible even when the terminal was set to be hidden in the metadata. After investigating, I realized that key parts of the metadata handling function were missing from the repository I pulled from.

I added a terminal property to the part schema as a boolean flag to address this. This flag controls whether the terminal should be shown or hidden for each tutorial part. While I couldn't fully verify the solution due to missing parts in the repository, I contributed this schema change as a potential fix to help manage terminal visibility.

Additions Made:

Added terminal property to the part schema to control terminal visibility for each part.
Introduced logic based on this flag to toggle the terminal visibility in the interface.

All added components have been commented out with a full description of functionality and purpose

### DIFF
--- a/packages/runtime/src/lesson-files.spec.ts
+++ b/packages/runtime/src/lesson-files.spec.ts
@@ -1,3 +1,4 @@
+// where is global?
 import { describe, test, expect, beforeAll, vi, afterAll, beforeEach } from 'vitest';
 import { LessonFilesFetcher } from './lesson-files.js';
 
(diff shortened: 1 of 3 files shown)


AriPerkkio (changes requested):
Thanks for the PR @SeanHersh!

However these changes don't look right to me. This bug fix should not require any changes in how metadata is configured. We'll also need test cases that verify the fix in order to accept PRs for this. 

It also seems that this issue might already be working just fine: https://github.com/stackblitz/tutorialkit/issues/300#issuecomment-2380657056.

PR #369, opened 2024-10-10
feat: add file extension based icons

Charan167:
- Adds files based on file extensions

This PR was made to get your feedback and make the necessary changes.


packages/react/src/core/FileTree.tsx
```diff
@@ -183,6 +183,7 @@ interface FileProps {
 }
 
 function File({ file: { depth, name }, onClick, selected }: FileProps) {
+  const fileIcon = name ? getFileTreeIcon(name) : '';
```
AriPerkkio, line 186: Overall looks good! 💯 

Couple of improvement ideas:

- Let's create a `getFileExtension` function that can be called with this `name`. 
  ```js
  function getFileExtension(filename: string) {
    const parts = filename.split('.');

    // drop the filename if there's no extension
    parts.shift();

    const extension = parts.at(-1) || '';

    return extension
  }
  ```

- Create a `extensionsToIcons` `Map` that holds all the extension-to-icon mappings. This should be created top-level somewhere between `interface Props` and `export function FileTree`.
  ```js
  const extensionsToIcons = new Map([
    ['ts', 'i-ph-file-ts-duotone'],

    ['cjs', 'i-ph-file-js-duotone'],
    ['mjs', 'i-ph-file-js-duotone'],
    ['js', 'i-ph-file-js-duotone'],

    ['html', 'i-ph-file-html-duotone'],

    ... and so on
  ]);
  ```

- Let's replace this line 186 with following:
  ```diff
  - const fileIcon = name ? getFileTreeIcon(name) : '';
  + const extension = getFileExtension(name);
  + const fileIcon = extensionsToIcons.get(extension) || 'i-ph-file-duotone';
  ```

AriPerkkio (changes requested):

packages/react/src/core/FileTree.tsx
AriPerkkio: Could you also create some examples files in `packages/template/src/content/tutorial/1-basics/1-introduction/2-foo/_files/src` so that we can easily test all of these extensions when opening `http://localhost:4321/1-basics/custom-chapter-slug/foo`.

Charan167: Hi there! I made the necessary changes to the code, but it seems the GitHub actions failed at first because I missed adding a newline before the return. I went ahead and edited the file directly on GitHub, but it's still not passing the actions. Could you kindly take a look at the code and let me know what might be wrong? 😅

packages/react/src/core/FileTree.tsx
```diff
@@ -183,6 +202,9 @@ interface FileProps {
 }
 
 function File({ file: { depth, name }, onClick, selected }: FileProps) {
+  const extension = getFileExtension(name);
+  const fileIcon = extensionsToIcons.get(extension) || 'i-ph-file-duotone';
+
```
AriPerkkio, line 207:
```suggestion

```

packages/react/src/core/FileTree.tsx
```diff
@@ -183,6 +202,8 @@ interface FileProps {
 }
 
 function File({ file: { depth, name }, onClick, selected }: FileProps) {
+  const extension = getFileExtension(name);
+  const fileIcon = extensionsToIcons.get(extension) || 'i-ph-file-duotone';
```
AriPerkkio, line 206:
```suggestion
  const fileIcon = extensionsToIcons.get(extension) || 'i-ph-file-duotone';

```

packages/react/src/core/FileTree.tsx
AriPerkkio: The changes look good! I'll check rest of this PR next week, thanks!

AriPerkkio (approved):
Awesome work once again @Charan167! Thanks for the PR! 🙌 

These changes look good to me. Any thoughts @sulco or @Nemikolh before we merge this? This is how the example page looks like:

image

The available icons can be found here https://icon-sets.iconify.design/ph/?query=file- just make sure to select `Duotone` filter:

image

Nemikolh:
@AriPerkkio, looks good! I think we can merge it! :smiley:

PR #372, opened 2024-10-13
feat(terminal): restrict default allowed commands to 'ls' and 'echo'

### LINKED ISSUE #302: `terminal.allowCommands` should have more strict default value

By default the terminal allows users to use any available command. In tutorials this is usually not intended experience. The lesson authors should define commands that are allowed - if any. 

Let's change the defaults value of `terminal.allowCommands` from `[]` to something like `['ls', 'echo']`. 

Lesson authors should still be able to allow all possible commands by specifying `terminal.allowCommands: []` in metadata.

Recommend authors to always specify `terminal.allowCommands`. 

### Additional context

_No response_

RonithManikonda:
Change the default value of `terminal.allowCommands` from an empty array to `['ls', 'echo']`. This change prevents users from running all available commands by default in tutorials, enhancing security and focus. Lesson authors can still allow all commands by specifying `terminal.allowCommands: []` in the metadata.

BREAKING CHANGE: The default value of `terminal.allowCommands` is now restricted to `['ls', 'echo']`. To allow all commands, explicitly set `terminal.allowCommands: []` in the metadata.

Closes #302

### DIFF
--- a/docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx
+++ b/docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx
@@ -312,7 +312,7 @@ You can set terminal open by default by specifying the `open` value.
 
 An interactive terminal will disable the output redirect syntax by default. For instance, you cannot create a file `world.txt` with the contents `hello` using the command `echo hello > world.txt`. The reason is that this could disrupt the lesson if a user overwrites certain files. To allow output redirection, you can change the behavior with the `allowRedirects` setting. You can define this setting either per panel or for all panels at once.
 
-Additionally, you may not want users to run arbitrary commands. For example, if you are creating a lesson about `vitest`, you could specify that the only command the user can run is `vitest` by providing a list of `allowCommands`. Any other command executed by the user will be blocked. You can define the `allowCommands` setting either per panel or for all panels at once.
+Additionally, you may not want users to run arbitrary commands. For example, if you are creating a lesson about vitest, you could specify that the only command the user can run is vitest by providing a list of allowCommands. By default, the allowed commands are ls and echo. Providing a list of allowCommands will override these defaults, and specifying an empty list will allow all commands. Any other command executed by the user will be blocked. You can define the allowCommands setting either per panel or for all panels at once.
 
 By default, in every new lesson terminals start a new session. If you want to keep the terminal session between lessons, you can specify the `id` property for a given terminal panel and keep the same `id` across lessons.
 <PropertyTable inherited type="Terminal" />
--- a/packages/runtime/src/webcontainer/terminal-config.ts
+++ b/packages/runtime/src/webcontainer/terminal-config.ts
@@ -200,6 +200,9 @@ export class TerminalPanel implements ITerminal {
   }
 }
 
+// set the default commands for the terminal
+const DEFAULT_COMMANDS = ['ls', 'echo'];
+
 /**
  * Normalize the provided configuration to a configuration which is easier to parse.
  *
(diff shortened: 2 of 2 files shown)


docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx
```diff
@@ -312,7 +312,7 @@ You can set terminal open by default by specifying the `open` value.
...
 
-Additionally, you may not want users to run arbitrary commands. For example, if you are creating a lesson about `vitest`, you could specify that the only command the user can run is `vitest` by providing a list of `allowCommands`. Any other command executed by the user will be blocked. You can define the `allowCommands` setting either per panel or for all panels at once.
+Additionally, you may not want users to run arbitrary commands. For example, if you are creating a lesson about vitest, you could specify that the only command the user can run is vitest by providing a list of allowCommands. By default, the allowed commands are ls and echo. Providing a list of allowCommands will override these defaults, and specifying an empty list will allow all commands. Any other command executed by the user will be blocked. You can define the allowCommands setting either per panel or for all panels at once.
```
AriPerkkio, line 315: Let's keep the backticks: 

```suggestion
Additionally, you may not want users to run arbitrary commands. For example, if you are creating a lesson about `vitest`, you could specify that the only command the user can run is `vitest` by providing a list of `allowCommands`. By default, the allowed commands are `ls` and `echo`. Providing a list of `allowCommands` will override these defaults, and specifying an empty list will allow all commands. Any other command executed by the user will be blocked. You can define the `allowCommands` setting either per panel or for all panels at once.
```

AriPerkkio (changes requested):
Great start, these changes look good to me! 💯 

Though there are some failing unit tests - could you take a look at those? You can run them with:

```sh
$ cd packages/runtime/
$ pnpm test
```

packages/runtime/src/webcontainer/terminal-config.spec.ts
```diff
@@ -144,13 +144,13 @@ describe('TerminalConfig', () => {
     expect(config.panels[0].title).toBe('TERM 1');
     expect(config.panels[0].processOptions).toEqual({
       allowRedirects: true,
-      allowCommands: ['echo'],
+      allowCommands: ['ls', 'echo'],
```
AriPerkkio, line 147: This (and change below) doesn't look right. Above we are setting `allowCommands: ['echo']`. 

```ts
134 |    const config = new TerminalConfig({
135 |      allowRedirects: true,
136 |      allowCommands: ['echo'],          // <--
```

Here on the assertion we should see just `echo`, no `ls`.

```suggestion
      allowCommands: ['echo'],
```

packages/runtime/src/webcontainer/terminal-config.ts
```diff
@@ -258,7 +273,7 @@ function normalizeTerminalConfig(config?: TerminalSchema): NormalizedTerminalCon
             id: panel.id,
             title: panel.title,
             allowRedirects: panel.allowRedirects ?? config.allowRedirects,
-            allowCommands: panel.allowCommands ?? config.allowCommands,
+            allowCommands: panel.allowCommands ?? DEFAULT_COMMANDS,
```
AriPerkkio, line 276: Here we should resolve `panel.allowCommands` values too. If they are not specified, default to the top level values `options.allowCommands`:

```suggestion
            allowCommands: panel.allowCommands ? resolveAllowCommands(panel.allowCommands) : options.allowCommands,
```

AriPerkkio (changes requested):

AriPerkkio (changes requested):
Some of the tests are still failing, could you take a look at those? 👀

AriPerkkio (approved):
Looks good, great work @RonithManikonda! 💯 

As this change is a breaking change, we'll merge this into `main` before releasing version 2.0.0 of TutorialKit.

PR #373, opened 2024-10-14
fix(cli): list eject command in help output

d3lm:
I noticed that `eject` wasn't listed when you use `--help`. This PR adds it to the output as well as fixes a type error.


packages/cli/src/commands/eject/index.ts
```diff
@@ -141,6 +142,8 @@ async function _eject(flags: EjectOptions) {
       }
     }
   }
+
+  return 0;
```
d3lm, line 146: This fixes a type error.

packages/cli/src/index.ts
```diff
@@ -44,6 +44,7 @@ async function runCommand(cmd: CLICommand, flags: yargs.Arguments): Promise<numb
         tables: {
           Commands: [
             ['create', 'Create new tutorial app'],
+            ['eject', 'Eject routes'],
```
d3lm, line 47: @AriPerkkio Not sure what you have in mind for the description here. I suppose it can be more descriptive and `Eject routes` also isn't entirely correct. Any thoughts?

Nemikolh: Maybe we could say that it exposes the astro pages and components that are normally handled by the integrations which gives full control about the Astro app. 

I suppose this is it, so how about:

```suggestion
            ['eject', 'Move all default pages and components in your project, providing full control over the Astro app']
```

d3lm: Yep, whatever you feel makes more sense 👍 Feel free to just push into this branch. I'd say `... into your project,...` tho.

AriPerkkio: Sounds good, let's use this.

packages/cli/src/commands/eject/index.ts
AriPerkkio: Good catch!

AriPerkkio (approved):
Looks good, thanks!

Issue #304, opened 2024-09-02
Parts of only one lesson


eric-burel:
### Is your feature request related to a problem?

I'd like to write intro/outro parts that may not need a 3-level subdivision between part, chapter and lesson.

Currently, I need multiple levels so my introduction page maybe `/getting-started/intro/about", instead of just `/getting-started`. The UX is not as smooth as it could when it comes to writing proper course introductions or conclusions.

### Describe the solution you'd like.

Allow having 2-levels or 1-level hierarchy.

```
intro
|_ _files
|_ content.md
|_ meta.md
short-part
|_ chapter-1
|__ _files
|__ content.md
```

### Describe alternatives you've considered.

This seems a little change but I understand how this can affect UX.
- We could allow having content in parts and chapters ? Maybe have special "intro.md"/"ending.md" files ?
- Or keep content only in type "lesson", but allow a lesson to live outside of a chapter or part ? In this approach, as a default, we write lessons in a tutorial. Optionally, we can move lessons to a chapter. And optionnaly, we can move chapters into parts.
- We could instead craft independant Astro pages, and focus on linking them to specific courses via meta data? For instance adding an "outro: /some-page" field to the chapter definition or something alike. But seems convoluted, mixing the mental model of Astro and the mental model of Tutorialkit.

### Additional context

The file `packages/astro/src/default/utils/routes.ts` seems to be responsible to route creation and could be tweaked quite easily. However such a change may have consequences on the mental model around parts/chapters/lessons in other places.

Nemikolh:
Thanks for opening this issue! 

I agree with the general idea proposed here, although I would instead allow the tutorial meta to have lessons or chapters which would achieve the same thing but keep the existing logic for parts and chapters.

If we don't allow mixed hierarchies where a tutorial has both parts and chapters / lessons, then it should be pretty easy to add.

PR #374, opened 2024-10-14
feat(astro): support lessons without parts or chapters

AriPerkkio:
- Closes https://github.com/stackblitz/tutorialkit/issues/304
- Example without chapters deployed at https://feat-remove-chapters.tutorial-vite-plugin.pages.dev
  

Adds support for defining lessons without chapters or parts. 

```
src/content/tutorial
│
├── meta.md # Tutorial metadata
│
└── lesson-1 # Lesson in root! No part or chapter.
   ├── content.md
   ├── _files
   └── _solution
```

```
src/content/tutorial
│
├── meta.md # Tutorial metadata
│
└── part-1
    └── lesson-1 # Lesson in part! No chapter.
        ├── content.md
        ├── _files
        └── _solution
```

Custom order can be defined using `lessons` in metadata:

```md
---
type: 'tutorial'
lessons: ['lesson-one', 'lesson-two']
---
```

```md
---
type: 'part'
lessons: ['lesson-one', 'lesson-two']
---
```

Mixing structures is not supported and an error is thrown when these are not followed: 

- A tutorial can have one-or-many `part`'s or one-or-many `lesson`'s
- A part can have one-or-many `chapter`'s or one-or-many `lesson`'s

In practice:

```
src/content/tutorial
├── meta.md # Tutorial
├── lesson-1 # Lesson
└── part-1 # Part

> Error: Cannot mix lessons and parts in a tutorial. Either remove the parts or move root level lessons into a part.

src/content/tutorial
├── meta.md # Tutorial
└── part-1 # Part
    ├── lesson-1 # Lesson
    └── chapter-1 # Chapter

> Error: Cannot mix lessons and chapters in a part. Either remove the chapter from 1-part or move the lessons into a chapter.
```

However you can have multiple `part`'s, where some contain just `chapter`'s and some contain just `lesson`'s:

```
src/content/tutorial
├── meta.md # Tutorial
│
├── part-1 # Part with lessons
│   └── lesson-1 # Lesson
│
└── part-2 # Part with chapters
    └── chapter-1 # Chapter
```

[tk-mixed-hierarchy.webm](https://github.com/user-attachments/assets/bd991110-24c8-4222-9eeb-8ffb5346f8b9)


packages/react/src/Nav.tsx
AriPerkkio, line 1: I've pretty much rewritten this whole component. I tried to re-use the old styling as much as possible but there can be something I've missed. Here's comparison of `demo.tutorialkit.dev` and this PR:

image

docs/tutorialkit.dev/src/content/docs/guides/creating-content.mdx
```diff
@@ -36,6 +37,64 @@ This structure is reflected in the directory structure of your TutorialKit proje
...
+                - content.md
+              - installing/
+            - project-structure/
+            - meta.md
+        - config.ts
+        - templates/
+    </FileTree>
+  </TabItem>
+</Tabs>
+
```
AriPerkkio, line 97: Renders to 
image

e2e/package.json
```diff
@@ -7,6 +7,10 @@
     "preview": "astro build && astro preview",
     "dev:override-components": "astro dev --config ./configs/override-components.ts",
     "preview:override-components": "astro build --config ./configs/override-components.ts && astro preview --config ./configs/override-components.ts",
+    "dev:lessons-in-root": "astro dev --config ./configs/lessons-in-root.ts",
+    "preview:lessons-in-root": "astro build --config ./configs/lessons-in-root.ts && astro preview --config ./configs/lessons-in-root.ts",
+    "dev:lessons-in-part": "astro dev --config ./configs/lessons-in-part.ts",
+    "preview:lessons-in-part": "astro build --config ./configs/lessons-in-part.ts && astro preview --config ./configs/lessons-in-part.ts",
```
AriPerkkio, line 13: I'm planning to create `scripts/start.mjs` that can be called like `node scripts/start.mjs --dev lessons-in-root` to reduce these verbose scripts.

packages/astro/src/default/utils/content.ts
```diff
@@ -75,250 +208,226 @@ export async function getTutorial(): Promise<Tutorial> {
...
-          logger.warn(`Could not find '${lessonId}', it won't be part of the chapter '${chapterId}'.`);
-          continue;
-        }
+  return slug;
+}
 
-        if (!chapter.firstLessonId) {
-          chapter.firstLessonId = lessonId;
-        }
+function resolveIds(
```
Nemikolh, line 297: Nice, this function is pretty cool, makes the logic focused in a single function.

packages/astro/src/default/utils/content.ts
```diff
@@ -75,250 +208,226 @@ export async function getTutorial(): Promise<Tutorial> {
...
+      if (partId && chapterId) {
+        if (!tutorial.parts[partId].chapters[chapterId]) {
+          throw new Error(`Could not find chapter '${chapterId}'`);
+        }
+
+        lesson.chapter = {
+          id: chapterId,
+          title: tutorial.parts[partId].chapters[chapterId].data.title,
+        };
+      }
```
Nemikolh, line 237: I'm wondering if we should do this in a second pass in case parts, chapters and lessons are not in order.

It's also something that could have been an issue with the previous logic though so maybe we don't need to care about this.

packages/astro/src/default/utils/content.ts
```diff
@@ -75,250 +208,226 @@ export async function getTutorial(): Promise<Tutorial> {
...
-      ),
-    };
+function sortTutorialLessons(tutorial: Tutorial, metadata: TutorialSchema) {
+  const lessonIds = tutorial.lessons.map((lesson) => lesson.id);
 
-    if (prevLesson) {
-      const partSlug = _tutorial.parts[prevLesson.part.id].slug;
-      const chapterSlug = _tutorial.parts[prevLesson.part.id].chapters[prevLesson.chapter.id].slug;
+  // lesson ID alone does not make a lesson unique - combination of lessonId + chapterId + partId does
+  const lessonOrder: { lessonId: Lesson['id']; chapterId?: Chapter['id']; partId?: Part['id'] }[] = [];
```
Nemikolh, line 372: Argh it's frustrating that we can't use a `Map` in there. In theory we should but because in JS it only work with reference identity for objects (as we can't provide custom hash / equals functions), the only option would be to use a string that can be guaranteed to be unique or use an array like you did which doesn't suffer from the uniqueness issue.

At first I was going to suggest to use `:` to join the three things but then given that there won't be many, I'm more comfortable with a robust solution that uses an array.

packages/react/src/Nav.tsx
```diff
@@ -95,123 +118,67 @@ export function Nav({ lesson: currentLesson, navList }: Props) {
...
-              </Accordion.Item>
-            </li>
-          );
-        })}
-      </Accordion.Root>
-    </ul>
+    <Accordion.Root asChild collapsible type="single" defaultValue={`${level}-${activeItems[level]}`}>
+      <ul className={classNames(className)}>
+        {items.map((item, index) => (
+          <NavListItem key={index} {...item} index={index} level={level} activeItems={activeItems} i18n={i18n} />
```
Nemikolh, line 132:
```suggestion
          <NavListItem key={item.id} {...item} index={index} level={level} activeItems={activeItems} i18n={i18n} />
```

packages/react/src/Nav.tsx
```diff
@@ -95,123 +118,67 @@ export function Nav({ lesson: currentLesson, navList }: Props) {
...
-        );
-      })}
-    </ul>
+        <Accordion.Content className={navStyles.AccordionContent}>
+          <NavListComponent
+            className="mt-1.5  pl-4.5"
+            items={sections}
+            activeItems={activeItems}
+            i18n={i18n}
+            level={1 + level}
```
Nemikolh, line 178: Super nitpick:

```suggestion
            level={level + 1}
```

packages/react/src/Nav.tsx
```diff
@@ -95,123 +118,67 @@ export function Nav({ lesson: currentLesson, navList }: Props) {
...
-            >
-              {lesson.title}
-            </a>
-          </li>
-        );
-      })}
-    </ul>
+        <Accordion.Content className={navStyles.AccordionContent}>
+          <NavListComponent
+            className="mt-1.5  pl-4.5"
```
Nemikolh, line 174:
```suggestion
            className="mt-1.5 pl-4.5"
```

Nemikolh (commented):
Fantastic work on this @AriPerkkio ! :tada: 

Can't wait to get this landed :smiley: 

I finished a first pass, I'll try to do another pass later this week :crossed_fingers:

e2e/test/navigation.lessons-in-part.test.ts
```diff
@@ -0,0 +1,19 @@
+import { test, expect } from '@playwright/test';
+
+test('user can navigate between lessons using breadcrumbs', async ({ page }) => {
+  await page.goto('/part-one/lesson-one');
+
+  await expect(page.getByRole('heading', { level: 1, name: 'Lessons in part test - Lesson one' })).toBeVisible();
+  await expect(page.getByText('Lesson in part without chapter')).toBeVisible();
+
+  // navigation select can take a while to hydrate on page load, click until responsive
+  await expect(async () => {
```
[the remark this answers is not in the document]
AriPerkkio: The button can become visible before it has been hydrated. Clicking it doesn't do anything. So here if the last `await expect(page.locator('[data-state="open"]', { has: button })).toBeVisible({ timeout: 50 })` fails, the whole callback inside `expect(callback).toPass()` will be retried -> button is clicked again.

This hydration issue might be related to running `<root>/e2e` in dev mode only. I'll try in follow-up PR that if we switch to `vite build --watch & vite preview`, would this issue disappear.

packages/astro/src/default/pages/index.astro
```diff
@@ -4,11 +4,13 @@ import { joinPaths } from '../utils/url';
 
 const tutorial = await getTutorial();
 
-const part = tutorial.parts[tutorial.firstPartId!];
-const chapter = part.chapters[part?.firstChapterId!];
-const lesson = chapter.lessons[chapter?.firstLessonId!];
+const lesson = tutorial.lessons[0];
```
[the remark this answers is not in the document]
AriPerkkio: Yes - we will always have lessons. Lessons may be _part of_ part or chapter. It's easier to always start to loop from Tutorial's lessons and then proceed checking if they have other levels.

Also all lessons at tutorial are sorted.

packages/astro/src/default/utils/content.ts
AriPerkkio: And this is intentionally verbose 😄

packages/astro/src/default/utils/content.ts
AriPerkkio: At this point they should be ordered:

https://github.com/stackblitz/tutorialkit/blob/0b2896a0e9e63a74ce8a3775523e9592e4b6b70f/packages/astro/src/default/utils/content.ts#L20-L22

https://github.com/stackblitz/tutorialkit/blob/0b2896a0e9e63a74ce8a3775523e9592e4b6b70f/packages/astro/src/default/utils/content.ts#L271-L278

But yes, it's not visible from this function only. Should we move the `sortCollection` function call inside this function, so that we can be sure it's always sorted before looping it?

packages/astro/src/default/utils/content.ts
AriPerkkio: I think I had `-` separated IDs as key at one point here. 😄 
But in the last part where `tutorial.lessons` are looped, it was much nicer to directly match to objects instead of splitting the IDs again.

This function is quite complex but only because we provide so many ways to construct tutorials and customize the ordering. But I ended up building extensive set of test cases for all ordering cases I could think of:

https://github.com/stackblitz/tutorialkit/blob/0b2896a0e9e63a74ce8a3775523e9592e4b6b70f/packages/astro/src/default/utils/content.spec.ts#L284-L543

Nemikolh (approved):
Amazing work on this @AriPerkkio! :tada:

Super impressed with how much cleaner the code is and overall this is a really high quality PR :muscle:

AriPerkkio:
Thanks for review @Nemikolh, let's get this merged! 💯

PR #377, opened 2024-10-16
fix: warn when using `.svg` in `meta.image`

eric-burel:
I figured the hard way that most social platforms won't actually support SVG open graph cover images, which might be confusing since we use the logo as the default cover image, and the default logo is SVG.

I've added a warning to detect this scenario, and added some documentation.

Maybe we could later improve the default setup to provide a default .png cover image directly? I've tested setting "meta.image" at tutorial level for all lessons and it works perfect, so maybe we could include a TutorialKit cover as "cover.png" and configure it directly there, rather than falling back on the logo.


packages/astro/src/default/components/MetaTags.astro
```diff
@@ -15,6 +15,10 @@ if (meta.image) {
   }
 }
 imageUrl ??= readLogoFile('logo', true);
+if (imageUrl?.endsWith('.svg')) {
+  console.warn(`Using a SVG open graph image "${imageUrl}". This is not supported by most social platforms. 
+  You should rather set "meta.image" to a raster image (PNG, WEBP).`);
+}
```
AriPerkkio, line 21: By default `npm create tutorial` creates `public/logo.svg`, which is used here by default when `meta.image` is not defined. This means that this error message would be shown to all new projects with default settings.

eric-burel: Yes I came to the same conclusion so that's why I think it would be reasonable to also add a "cover.webp" in "/public", and use that as the default
Having no image is a major penalty on social media and search engines imo, having at least something like "A tutorialkit lesson" with the framework logo would be totally acceptable and preferable to nothing.

docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx
```diff
@@ -404,6 +404,25 @@ type TemplateType = "html" | "node" | "angular-cli" | "create-react-app" | "java
...
+meta:
+  image: /cover.png
+  title: Title shown on social media and search engines
+  description: Description shown on social media and search engines
+```
+
+TutorialKit will use your logo as the default image.
+If your logo uses the SVG format, it may not display on most social platforms.
+Use a raster format instead, such as WEBP or PNG.
+
```
AriPerkkio, line 425: Could you create separate PR for these, so that missing default og-image would not block documentation updates?

packages/astro/src/default/components/MetaTags.astro
AriPerkkio: @sulco do you have any good images that we could use as default og-image in the created tutorials?

docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx
AriPerkkio: Moved here: https://github.com/stackblitz/tutorialkit/pull/379

AriPerkkio, line 425:
```suggestion
```

AriPerkkio:
Let's host a static png/jpg on `tutorialkit.dev` that used by default, instead of the default logo.svg. But let's also include this warning message from this PR.

PR #378, opened 2024-10-16
feat(astro): add `custom` configuration option for passing custom fields

eric-burel:
All in the title:  a `custom` record on each configuration unlocks user-land features where we need to associate metadata to a specific entry:
- building a table of content
- building an RSS feed
- improved sitemap
- a tag system
- access extra meta information when crafting custom components : displaying the author name, publication date or such

It should be future proof in the sense that even if a feature is later brought to TutorialKit core, users will just remove their custom code when updating, but their app won't break.

Having it inherited could be relevant for some custom fields (author name), weird for others (publication date) : I feel that inheritance is still more convenient than setting custom fields on every tutorial? Also will inheritance proceed to a deep merge or not?

### DIFF
--- a/packages/types/src/schemas/common.ts
+++ b/packages/types/src/schemas/common.ts
@@ -203,12 +203,17 @@ export const editorSchema = z.union([
   }),
 ]);
 
+const customSchema = z.record(z.string(), z.any());
+
 export type TerminalPanelType = z.infer<typeof panelTypeSchema>;
 export type TerminalSchema = z.infer<typeof terminalSchema>;
 export type EditorSchema = z.infer<typeof editorSchema>;
+export type CustomSchema = z.infer<typeof customSchema>;
 
 export const webcontainerSchema = commandsSchema.extend({
   meta: metaTagsSchema.optional(),
+
+  custom: customSchema.optional(),
 
   previews: previewSchema
     .optional()
--- a/packages/types/src/schemas/common.ts
+++ b/packages/types/src/schemas/common.ts
@@ -213,7 +213,7 @@ export type CustomSchema = z.infer<typeof customSchema>;
 export const webcontainerSchema = commandsSchema.extend({
   meta: metaTagsSchema.optional(),
 
-  custom: customSchema.optional(),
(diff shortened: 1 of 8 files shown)


AriPerkkio (approved):
Looks good to me, great work! 💯 

I've added one test case to verify the implementation.

@Nemikolh do you see any pitfalls in this feature? If not, we can merge this.

eric-burel:
Thanks! For the record I think the alternative would be extending the "tutorial" collection, allowing more fields in it upon config. The main advantage is that it would be typesafe (but maybe it can also be done with "custom" using an ambient type somehow), and maybe extending a collection may have more benefit than just adding fields (though I don't know that). But it feels more complex while adding a "custom" dict is trivial codewise and should be far enough for the listed use cases.

AriPerkkio:
> extending the "tutorial" collection, allowing more fields in it upon config

Allowing users to add custom fields in root of the metadata could cause conflicts with possible future built-in metadata fields. Having this separate `custom` option makes sure we don't introduce new fields that accidentally overrides user's own custom fields.

packages/types/src/schemas/common.ts
```diff
@@ -203,13 +203,18 @@ export const editorSchema = z.union([
...
+
 export type TerminalPanelType = z.infer<typeof panelTypeSchema>;
 export type TerminalSchema = z.infer<typeof terminalSchema>;
 export type EditorSchema = z.infer<typeof editorSchema>;
+export type CustomSchema = z.infer<typeof customSchema>;
 
 export const webcontainerSchema = commandsSchema.extend({
   meta: metaTagsSchema.optional(),
 
+  custom: customSchema.optional(),
```
Nemikolh, line 216: Oh I think we could add a `.describe` to explain what this is for (it shows up in the extension after).

Nemikolh (approved):
This is really nice! 

Just realised that this can also be used at runtime by a custom component. The nice thing about using this vs just having a props is that it's inherited.

Maybe we could mention it in the docs?

PR #389, opened 2024-10-22
fix(deps): move `react` & `react-dom` to peer dependencies

AriPerkkio:
Let's see if this fixes "Invalid hook call" errors caused when custom components are used in mdx-files.


AriPerkkio:
Nope, did not help.

Issue #396, opened 2024-10-27
Lessons with only the terminal and no browser


noam-honig:
### Is your feature request related to a problem?

Some of the lessons in my tutorial are about unit testing, where I do not need a ui - everything happens in the terminal

### Describe the solution you'd like.

It'll be great if I could set `preview` false and `terminal` open so that'll the configuration for that lesson

### Describe alternatives you've considered.

I wrote an instruction for the student to open the terminal and make the preview as small as possible - it's not so cool, as it keeps showing the progress indicator of server starting - here it is in my tutorial:
![image](https://github.com/user-attachments/assets/668ff65e-acc8-4201-89c7-e2042aa5fa92)

You can try it yourself at:
https://learn.remult.dev/in-depth/9-testing/1-introduction

AriPerkkio:
You can do this already. Here's live example and sources for it:
- https://tutorial-vite-plugin.pages.dev/vite-plugin/env-plugin/env-vars-in-node/
- https://github.com/AriPerkkio/tutorial-vite-plugin/blob/main/src/content/tutorial/vite-plugin/env-plugin/env-vars-in-node/content.md?plain=1#L1-L13

noam-honig:
Thanks - appreceate your support

PR #397, opened 2024-10-28
docs: add `previews: false` example

noam-honig:
Based on the answer to issue #396 

https://github.com/stackblitz/tutorialkit/issues/396#issuecomment-2440130125


AriPerkkio (changes requested):

AriPerkkio (approved):
Thanks once again @noam-honig! 💯

PR #398, opened 2024-10-28
access current lesson from Astro.locals

eric-burel:


AriPerkkio (commented):
I think this is currently possible via TutorialKit APIs: https://tutorialkit.dev/guides/how-to-use-tutorialkit-api/

You could create a custom component that accesses the lesson:

```jsx

export function CustomComponent() {
  const lesson = tutorialStore.lesson;

  console.log(lesson);

  return <></>;
}
```

eric-burel:
Alright ! So the TutorialStore is a stateful singleton as far as I can tell from the code.

However I see `setLesson` being called only in `WorkspacePanelWrapper` in a `useEffect` hook: is this value available in server-side Astro components as well? I feel like it only works in islands, and is strongly tied to the WorkspacePanel component loading first.

PR #399, opened 2024-10-30
fix(react): add vertical scroll bar to editor file tree

isaacplmann:
Small ui fix that adds `overflow-y: auto` to the editor file tree for when there are more files than fit in the editor panel

Before:
Screenshot 2024-10-30 at 7 07 12 PM

After:
Screenshot 2024-10-30 at 7 06 15 PM

### DIFF
--- a/packages/react/src/Panels/EditorPanel.tsx
+++ b/packages/react/src/Panels/EditorPanel.tsx
@@ -79,7 +79,7 @@ export function EditorPanel({
           </div>
         </div>
         <FileTree
-          className="flex-grow py-2 border-r border-tk-elements-app-borderColor text-sm overflow-y-auto overflow-x-hidden"
+          className="flex flex-col flex-grow py-2 border-r border-tk-elements-app-borderColor text-sm overflow-y-auto overflow-x-hidden"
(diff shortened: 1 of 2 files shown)


packages/react/src/Panels/EditorPanel.tsx
```diff
@@ -79,7 +79,7 @@ export function EditorPanel({
           </div>
         </div>
         <FileTree
-          className="flex-grow py-2 border-r border-tk-elements-app-borderColor text-sm"
+          className="flex-grow py-2 border-r border-tk-elements-app-borderColor text-sm overflow-y-auto overflow-x-hidden"
```
AriPerkkio, line 82: This makes context menu's trigger area grow quite much now. This is visible in dev mode's `http://localhost:4321/1-basics/custom-chapter-slug/foo`:

image

We'll need to add here

```suggestion
          className="flex flex-col flex-grow py-2 border-r border-tk-elements-app-borderColor text-sm overflow-y-auto overflow-x-hidden"
```

AriPerkkio (approved):
Looks good, thanks @isaacplmann! 💯

Issue #166, opened 2024-07-24
Easy download of the lesson code merging template + current files


eric-burel:
### Is your feature request related to a problem?

I am crafting a tutorial, and debugging the template application can be a bit tricky. The webcontainer loading time is acceptable for an application that works, but quite slow for a debugging process when it doesn't. Also when modifying the content of "_files" and "_solution" folder, the files may miss some context from the parent template and end up with bad typings (maybe there are solutions to that though), it's easier to work on a complete application. Finally end user may want to do a local run of the code.

### Describe the solution you'd like.

Be able to download the full template + _file/_solution code.

### Describe alternatives you've considered.

Created a dirty copy script with `cp`

```
# Copy an app to test it locally
mkdir -p "./tmp"
cp -R ./src/templates/nextjs/. ./tmp
cp -R ./src/content/tutorial/2-nextjs/2-various/1-streaming/_files/. ./tmp
```
Then I can debug the app in "tmp", finalize the code, and move it back to the solution.

It does the job making this issue not really dramatic, but we could imagine something built-in, all the more that you probably have merging logic in order to bootup the webcontainer (I didn't dig the codebase yet).

sulco:
Love this idea! We've been also discussing an "open in StackBlitz" functionality — that would allow e.g. for easy reproductions for bugs in lessons, continuing with the lessons, etc — and it looks like both features would benefit from the same ground work (like exposing the fs state to the developer)

AriPerkkio:
@eric-burel does https://github.com/stackblitz/tutorialkit/pull/219 solve this issue?

You can easily open the current status of TutorialKit's code editor + template in StackBlitz by clicking the button on top bar.

eric-burel:
Thanks that's perfect for a first iteration! I'd be eager to have a quick way to run a local install too, because at some point you often want learners to be able to run an app on their own computer (that's probably doable via Stackblitz UI but still need 2 actions instead of one). Could also be interesting to be able to see the whole codebase of the template directly within tutorialkit: this way you don't have to boot the webcontainer twice. Anyway I'll close this specific issue as it solves the main issue: seeing the whole app codebase.

eric-burel:
Hi folks, reopening this issue with new insights, there are legitimate scenarios where running the preview is not a good idea and running on StackBlitz is not sufficient. 

[...]

PR #415, opened 2024-11-13
feat(astro): add "Download lesson as zip" button

AriPerkkio:
- Closes https://github.com/stackblitz/tutorialkit/issues/166

### DIFF
--- a/docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx
+++ b/docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx
@@ -412,6 +412,19 @@ type TemplateType = "html" | "node" | "angular-cli" | "create-react-app" | "java
 
 ```
 
+### `downloadAsZip`
+Display a button for downloading the current lesson as `.zip` file.
+<PropertyTable inherited type="DownloadAsZip" />
+
+The `DownloadAsZip` type has the following shape:
+
+```ts
+type DownloadAsZip =
+    | boolean
+    | { filename?: string }
+
+```
+
 ##### `meta`
 
 Configures `<meta>` tags for Open Graph protocole and Twitter.
--- a/docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx
+++ b/docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx
@@ -413,7 +413,8 @@ type TemplateType = "html" | "node" | "angular-cli" | "create-react-app" | "java
 ```
 
 ### `downloadAsZip`
-Display a button for downloading the current lesson as `.zip` file.
+Display a button for downloading the current lesson as `.zip` file. Defaults to `false`.
+The default filename is constructed by concatenating folder names of part,  chapter and lesson.
 <PropertyTable inherited type="DownloadAsZip" />
 
 The `DownloadAsZip` type has the following shape:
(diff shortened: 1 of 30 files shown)


packages/astro/package.json
```diff
@@ -45,7 +45,7 @@
     "@tutorialkit/types": "workspace:*",
     "@types/react": "^18.3.3",
     "@unocss/reset": "^0.62.2",
-    "@webcontainer/api": "1.2.4",
+    "@webcontainer/api": "1.5.0",
```
AriPerkkio, line 48: TODO: Requires unreleased `1.5.1`.

packages/astro/src/default/components/DownloadButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+    </button>
+  );
+}
+
+async function onClick() {
+  const lesson = tutorialStore.lesson;
+
+  if (!lesson) {
+    throw new Error('Missing lesson');
+  }
```
Nemikolh, line 20: Oh I suppose we are throwing here because this can only happen in development if there's a bug? :eyes:

packages/astro/src/default/utils/content.ts
```diff
@@ -30,6 +30,7 @@ export async function getTutorial(): Promise<Tutorial> {
       tutorialMetaData.template ??= 'default';
       tutorialMetaData.i18n = Object.assign({ ...DEFAULT_LOCALIZATION }, tutorialMetaData.i18n);
       tutorialMetaData.openInStackBlitz ??= true;
+      tutorialMetaData.downloadAsZip ??= true;
```
Nemikolh, line 33: Oh I had an idea, what if instead of defaulting to `true` here which would add it to existing project that do an upgrade, we add the `downloadAsZip: true` attribute to the tutorial metadata for new project. So here:

https://github.com/stackblitz/tutorialkit/blob/5a2a64f35be823c7451af898e4bb6f91693fabff/packages/cli/overwrites/src/content/tutorial/meta.md?plain=1#L1-L6

What do you think?

docs/tutorialkit.dev/src/content/docs/reference/configuration.mdx
```diff
@@ -412,6 +412,19 @@ type TemplateType = "html" | "node" | "angular-cli" | "create-react-app" | "java
 
 ```
 
+### `downloadAsZip`
+Display a button for downloading the current lesson as `.zip` file.
+<PropertyTable inherited type="DownloadAsZip" />
+
```
Nemikolh, line 418: Maybe we should specify what the logic behind the default file name is?

Nemikolh (commented):
As always fantastic work on this! :star_struck:  Love the `onClick` handler at the module level :pinched_fingers:  

Only minor feedback!

packages/astro/package.json
Nemikolh: It's there :raised_hands:

packages/astro/src/default/components/DownloadButton.tsx
AriPerkkio: This is only to make Typescript happy 😄

packages/astro/src/default/utils/content.ts
AriPerkkio: Yup sounds good! Let's default to `false` and configure new projects to use it.

Nemikolh (approved):
Nice!! :star_struck:

PR #423, opened 2024-11-21
feat(dev): add support for footer override

### LINKED ISSUE #323: Overriding the footer

The webcontainer link text customization is not enough, because this footer is also the place where we may want to include legal terms, contact info and copyrights, especially when deploying a tutorial as a completely standalone website with no other content page.

Similarly to TopBar, customizing the footer.

Customizing "webcontainerLinkText" to feed it more content is not enough, as it will be wrapped by an anchor tag.

### Additional context

I am digging the code to better understand the implementation, and maybe contribute : I see that overidable components are typed in `packages/astro/src/default/env-default.d.ts`, which then allows to get them from a virtual import `packages/astro/src/default/components/TopBarWrapper.astro`.
I think this system is fine, so it could be expanded to more components. Coupled with slots, it's really great because you only need to provide the relevant slots in your own version to keep similar content around.

RonithManikonda:
adds support for overriding the footer component, similar to the TopBar override setup. Users can now customize footer content such as legal terms, contact info, and copyrights.

this change introduces `Footer.astro` and `FooterWrapper.astro` files to allow flexible slotting of footer items, improving customization for standalone tutorial deployments

close #323

### DIFF
--- a/packages/astro/src/default/components/Footer.astro
+++ b/packages/astro/src/default/components/Footer.astro
@@ -1,0 +1,13 @@
+<nav
+  class="bg-tk-elements-footer-backgroundColor border-t border-tk-elements-app-borderColor flex justify-between p-4"
+>
+  <div class="flex-1">
+    <slot name="terms" />
+  </div>
+  <div class="flex-1">
+    <slot name="contact-info" />
+  </div>
+  <div class="flex-1">
+    <slot name="copyright" />
+  </div>
+</nav>
--- a/packages/astro/src/default/components/FooterWrapper.astro
+++ b/packages/astro/src/default/components/FooterWrapper.astro
@@ -1,0 +1,14 @@
+---
+// FooterWrapper.astro
+import { Footer } from 'tutorialkit:override-components';
+import Terms from './Terms.astro';
+import ContactInfo from './ContactInfo.astro';
+
+const currentYear = new Date().getFullYear();
+---
+
+<Footer>
+  <Terms slot="terms" />
+  <ContactInfo slot="contact-info" />
+  <p slot="copyright">© {currentYear} Your Company. All rights reserved.</p>
+</Footer>


AriPerkkio (changes requested):
Hey @RonithManikonda, looks good so far! 💯 

There are still couple of places we need to update:

- Add the entrypoint here:

https://github.com/stackblitz/tutorialkit/blob/2740bc980798ca3582acc9fd35e5b3150e8d300d/packages/astro/package.json#L16-L25

- Mention the component here in the documentation:

https://github.com/stackblitz/tutorialkit/blob/04ea97faf0a142d3bc1edd7ab8aed89021f1558b/docs/tutorialkit.dev/src/content/docs/guides/overriding-components.mdx?plain=1#L12

- Add types here:

https://github.com/stackblitz/tutorialkit/blob/04ea97faf0a142d3bc1edd7ab8aed89021f1558b/packages/astro/src/default/env-default.d.ts#L10-L15

- Add component here:

https://github.com/stackblitz/tutorialkit/blob/2740bc980798ca3582acc9fd35e5b3150e8d300d/packages/astro/src/vite-plugins/override-components.ts#L1-L129

AriPerkkio (changes requested):
The `FooterWrapper.astro` should have following elements:

https://github.com/stackblitz/tutorialkit/blob/0e7cf3c082d4ceb298668e06a4914253327eb0c3/packages/astro/src/default/components/TutorialContent.astro#L19-L47

These should be passed to `Footer` with following named slots:
- `edit-page-link`
- `lesson-link`
- `webcontainers-link`

So for example the Edit Page -link becomes something like

```diff
-<div class="pb-4 mt-8 border-b border-tk-border-secondary">
+<div slot="edit-page-link" class="pb-4 mt-8 border-b border-tk-border-secondary">
  <a
    href={editPageLink}
    class="inline-flex flex-items-center text-tk-elements-link-secondaryColor hover:text-tk-elements-link-secondaryColorHover hover:underline"
```

Then in `Footer.astro` we would have those slots like:

```astro
<footer>
  <slot name="edit-page-link" />
  <slot name="lesson-links" />
  <slot name="webcontainers-link" />
</footer>
```

Then the `FooterWrapper.astro` should be used in `TutorialContent.astro`.

PR #437, opened 2025-01-27
feat(astro): add sensible default canonical url

eric-burel:
I got deindexed by Google due to the lack of a proper canonical link, so here is a small update to set it up when the website has a configured absolute URL.

image

The PR just needs a double check because I was running the demo code which doesn't have an astro.config so no Astro.site.

Current palliative is to setup a custom HeadTags.


Nemikolh (approved):
Nice! Thanks for the fix 🎉

PR #440, opened 2025-02-05
feat: add hosting config files

RonithManikonda:


AriPerkkio (changes requested):
Great start @RonithManikonda! 💯 

Some comments below. Also the test are currently stuck as there is no default value for this new option. The tests are using `--defaults` which skips all prompts and uses default values. You can test this CLI manually by running:

```sh
$ cd packages/cli
$ pnpm build
$ node ./dist/index.js create example

# Or to see how tests are using this, see packages/cli/tests/create-tutorial.test.ts
$ node ./dist/index.js create example --no-git --no-install --no-start --defaults
```

packages/cli/src/commands/create/index.ts
```diff
@@ -143,7 +144,20 @@ async function _createTutorial(flags: CreateOptions): Promise<undefined> {
...
+  const provider = await prompts.select({
+    message: 'Select hosting providers for automatic configuration:',
+    options: [
+      { value: 'Vercel', label: 'Vercel' },
+      { value: 'Netlify', label: 'Netlify' },
+      { value: 'Cloudflare', label: 'Cloudflare' },
+      { value: 'skip', label: 'Skip hosting configuration' },
+    ],
+    initialValue: 'Vercel',
+  });
```
AriPerkkio, line 156: The test cases are stuck as this prompt is waiting for selection. We need to add CLI flag that can be passed on by pass this, e.g. `--no-provider`.

Similar to `--git` and `--no-git`:

https://github.com/stackblitz/tutorialkit/blob/b2384ffb7ff8ce24f42dafe6d6c47e4fcf0d2384/packages/cli/src/commands/create/git.ts#L12-L18

https://github.com/stackblitz/tutorialkit/blob/b2384ffb7ff8ce24f42dafe6d6c47e4fcf0d2384/packages/cli/src/commands/create/index.ts#L31

Then pass it here in tests:

https://github.com/stackblitz/tutorialkit/blob/b2384ffb7ff8ce24f42dafe6d6c47e4fcf0d2384/packages/cli/tests/create-tutorial.test.ts#L43

packages/cli/src/commands/create/generate-hosting-config.ts
```diff
@@ -0,0 +1,80 @@
...
+        { value: 'Vercel', label: 'Vercel' },
+        { value: 'Netlify', label: 'Netlify' },
+        { value: 'Cloudflare', label: 'Cloudflare' },
+        { value: 'skip', label: 'Skip hosting configuration' },
+      ],
+      initialValue: DEFAULT_VALUES.provider,
+    })) as string;
+  }
+
+  if (provider === 'skip') {
```
AriPerkkio, line 28:
```suggestion
  if (!provider || provider === 'skip') {
```

packages/cli/src/commands/create/index.ts
```diff
@@ -261,7 +268,12 @@ function updatePackageJson(dest: string, projectName: string, flags: CreateOptio
   updateWorkspaceVersions(pkgJson.dependencies, TUTORIALKIT_VERSION);
   updateWorkspaceVersions(pkgJson.devDependencies, TUTORIALKIT_VERSION);
 
-  fs.writeFileSync(pkgPath, JSON.stringify(pkgJson, undefined, 2));
+  if (provider === 'Netlify' || provider === 'Cloudflare') {
```
AriPerkkio, line 271: This is only for Cloudflare: https://tutorialkit.dev/guides/deployment/#headers-configuration

```suggestion
  if (provider.toLowerCase() === 'cloudflare') {
```

packages/cli/src/commands/create/generate-hosting-config.ts
```diff
@@ -0,0 +1,80 @@
...
+      config = netlifyConfigRaw;
+      filename = 'netlify.toml';
+      break;
+    }
+    case 'Cloudflare': {
+      config = cloudflareConfigRaw;
+      filename = '_headers';
+      break;
+    }
+  }
```
AriPerkkio, line 63: Let's make this case insensitive so that CLI flags like `--provider netlify` work too.

```suggestion
  switch (provider.toLowerCase()) {
    case 'vercel': {
      config = typeof vercelConfigRaw === 'string' ? vercelConfigRaw : JSON.stringify(vercelConfigRaw, null, 2);
      filename = 'vercel.json';
      break;
    }
    case 'netlify': {
      config = netlifyConfigRaw;
      filename = 'netlify.toml';
      break;
    }
    case 'cloudflare': {
      config = cloudflareConfigRaw;
      filename = '_headers';
      break;
    }
  }
```

AriPerkkio (changes requested):
Tests and build are now passing, nice!

I've added couple of notes below. In addition to those, let's add some test cases in `packages/cli/tests/create-tutorial.test.ts`. You can run these tests with `pnpm run test` when in `packages/cli` directory.

For each provider, add a test case that builds the project and checks that it generated the expected hosting file. Here's and example for Netlify:

```ts
test('create a project with Netlify as provider', async (context) => {
  const name = context.task.id;
  const dest = path.join(tmpDir, name);

  await execa('node', [cli, 'create', name, '--no-install', '--no-git', '--defaults', '--provider', 'netlify'], {
    cwd: tmpDir,
  });

  const projectFiles = await fs.readdir(dest, { recursive: true });
  expect(projectFiles).toContain('netlify.toml');
});
```

And for Cloudflare's case we'll also need to check that `package.json` contains the `postbuild` script:

```ts
const packageJson = await fs.readFile(`${dest}/package.json`, 'utf8');
const json = JSON.parse(packageJson);

expect(json).toHaveProperty('scripts');
expect(json.scripts).toHaveProperty('postbuild');
expect(json.scripts.postbuild).toBe('cp _headers ./dist/');
```

packages/cli/src/commands/create/index.ts
```diff
@@ -143,7 +148,9 @@ async function _createTutorial(flags: CreateOptions): Promise<undefined> {
 
   await copyTemplate(resolvedDest, flags);
 
-  updatePackageJson(resolvedDest, tutorialName, flags);
+  const provider = await generateHostingConfig(resolvedDest, flags);
+
+  updatePackageJson(resolvedDest, tutorialName, flags, String(provider));
```
AriPerkkio, line 153:
```suggestion
  updatePackageJson(resolvedDest, tutorialName, flags, provider);
```

AriPerkkio (approved):

AriPerkkio:
Looks good, great work @RonithManikonda!

Let's merge this and I'll publish a new release including this feature.

PR #444, opened 2025-04-11
feat(astro): custom expressive code plugins

isaacplmann:
I would like to add custom expressive code plugins to change the way code blocks are rendered.

This PR adds an options `expressiveCodePlugins` property on the `tutorialkit` plugin that is instantiated in the `astro.config.ts`.

### DIFF
--- a/packages/astro/src/index.ts
+++ b/packages/astro/src/index.ts
@@ -1,5 +1,6 @@
 import { fileURLToPath } from 'node:url';
 import type { AstroConfig, AstroIntegration } from 'astro';
+import type { ExpressiveCodePlugin } from 'astro-expressive-code';
 import { extraIntegrations } from './integrations.js';
 import { updateMarkdownConfig } from './remark/index.js';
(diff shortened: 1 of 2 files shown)


AriPerkkio (approved):
Thanks, looks good! 💯 

I'll got to check why the e2e tests are breaking. Looks like unzipping the downloaded project fails now. Probably related to recent Webcontainer Node 20 upgrades (cc. @SamVerschueren) 🤔

SamVerschueren:
@AriPerkkio I tested the same code on StackBlitz and it seems to work fine (https://stackblitz.com/~/edit/node-wpdcdazw?file=index.js). The zip I used is downloaded from Bolt and thus uses the same export functionality as TutorialKit. So not entirely sure what goes wrong...

AriPerkkio:
> I'll got to check why the e2e tests are breaking. Looks like unzipping the downloaded project fails now. 

Created https://github.com/stackblitz/tutorialkit/issues/447 for this.

PR #445, opened 2025-04-11
fix(react): allow preview panel to be fully collapsed

isaacplmann:
My tutorial does not heavily use the preview panel, but always uses the terminal. This change allows the terminal to be open and the preview panel to be collapsed all the way (by the user).  The initial size of the preview panel stays the same.

### DIFF
--- a/packages/react/src/Panels/WorkspacePanel.tsx
+++ b/packages/react/src/Panels/WorkspacePanel.tsx
@@ -241,7 +241,7 @@ function PreviewsSection({
     <Panel
       id={hasPreviews ? 'previews-opened' : 'previews-closed'}
       defaultSize={hasPreviews ? 50 : 0}
-      minSize={10}
+      minSize={3}
       maxSize={hasPreviews ? 100 : 0}
       collapsible={!hasPreviews}
       className={classNames({
--- a/packages/react/src/Panels/WorkspacePanel.tsx
+++ b/packages/react/src/Panels/WorkspacePanel.tsx
@@ -241,7 +241,7 @@ function PreviewsSection({
   const [panelMinSize, setPanelMinSize] = useState(10);
 
   useEffect(() => {
-    const panelGroup: any = document.querySelector('[data-panel-group-id="right-panel-group"]');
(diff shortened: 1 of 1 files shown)


isaacplmann:
Seems like the multiple preview window e2e test is flaky. Even locally it will pass and not pass on subsequent runs without changing the code.

AriPerkkio:
@isaacplmann could you show screenshots of before-after this change?

I haven't seen that test case being flaky before. Could it be that it becomes flaky with this change when window is smalle enough?

isaacplmann:
When you click `Toggle Terminal` this is what the preview panel looks like both before and after the change.
Screenshot 2025-04-14 at 8 41 15 AM

Before the change, this is the smallest you can make the preview panel:
Screenshot 2025-04-14 at 8 41 37 AM

After the change, you can completely hide the preview panel with the terminal:
Screenshot 2025-04-14 at 8 40 03 AM

Here's an image of the e2e test passing locally:
Screenshot 2025-04-14 at 8 38 34 AM

isaacplmann:
Here's a screenshot of the `user can see multiple preview tabs` e2e test failing on `main`
Screenshot 2025-04-14 at 9 07 50 AM

packages/react/src/Panels/WorkspacePanel.tsx
```diff
@@ -241,7 +241,7 @@ function PreviewsSection({
     <Panel
       id={hasPreviews ? 'previews-opened' : 'previews-closed'}
       defaultSize={hasPreviews ? 50 : 0}
-      minSize={10}
+      minSize={3}
```
AriPerkkio, line 244: This `3` here is relative %-unit. On smaller screens this looks like:

image

I think we need to calculate the `minSize` based on window height here.

isaacplmann: Good catch.  That makes this PR a little more complex. I pushed changes that will calculate the percentage height based on a static pixel value

AriPerkkio (approved):
Looks good, thanks! 💯

PR #446, opened 2025-04-11
feat(astro): preserve file path for imported file code blocks

isaacplmann:
I would like to write an expressive code plugin that behaves differently for code blocks that were imported from the file system.  However, the import file logic removes all information about the file path before the expressive code plugins are called.

This change preserves that information so that expressive code plugins can access it.

### DIFF
--- a/packages/astro/src/remark/import-file.ts
+++ b/packages/astro/src/remark/import-file.ts
@@ -48,6 +48,7 @@ export function remarkImportFilePlugin(options: RemarkImportFilePluginOptions) {
 
           if (content) {
             node.value = content;
+            node.meta ||= '';
             node.meta += ' path=' + node.lang;
             node.lang = path.extname(relativeFilePath).slice(1);
           }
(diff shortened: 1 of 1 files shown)


packages/astro/src/remark/import-file.ts
```diff
@@ -48,6 +48,7 @@ export function remarkImportFilePlugin(options: RemarkImportFilePluginOptions) {
 
           if (content) {
             node.value = content;
+            node.meta += ' path=' + node.lang;
```
AriPerkkio, line 51: I'm not that familiar with expressive code's `node.meta`, but looking at the typings this can be `null | undefined`. We should probably do

```diff
+node.meta ||= '';
node.meta += ' path=' + node.lang;
```

... to avoid getting `'undefined path=somelang'`.

isaacplmann: good call. updated

AriPerkkio (approved):

PR #451, opened 2025-04-16
fix: headers not matching the documentation

geromegrignon:
## Description

The generated `vercel.json` file content includes wrong headers, not matching the ones from the documentation:

- there is a typo for the first header
- the key of the second one is copied from the first one


AriPerkkio (approved):
Looks good, thanks!

PR #454, opened 2025-06-08
fix: use `meta.title` as default `<title>`

eric-burel:
Hey folks, long time no see!

After reviewing my sites SEO, I've noticed that the titles were not properly configured. Because TutorialKit also uses the title for navigation, it injects the chapter and part title too.

To sum it up the proposal
- if meta.title is set, use it in the `<title>` tag: it will be picked up by search engines. But also visible in the browser window, so we lose the display of the part and chapter.
- if not set, the `<title>` tag is just the navigational title `part / chapter /lesson`

Related but not included in this PR: I suspect `trailingSlash: "never"` is a reasonable default in the config, otherwise search engines might pick pages as duplicate. Might be worth an investigation.

### DIFF
--- a/packages/astro/src/default/layouts/Layout.astro
+++ b/packages/astro/src/default/layouts/Layout.astro
@@ -18,7 +18,7 @@ const canonicalUrl = Astro.site ? new URL(Astro.url.pathname, Astro.site).toStri
 <html lang="en" transition:animate="none" class="h-full overflow-hidden">
   <head>
     <HeadTags>
-      <title slot="title">{title}</title>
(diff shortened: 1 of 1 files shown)


AriPerkkio (approved):
Yup, this makes sense. The default `title` is constructed from multiple parts, so it can indeed cause similar titles for all pages.

https://github.com/stackblitz/tutorialkit/blob/3c608fb92c9499fc1bf9902bd61a970f7ebcb29b/packages/astro/src/default/utils/routes.ts#L18

Thanks for the PR! 💯

PR #469, opened 2025-11-26
docs: remove extra verb and correct pronoun

brandonzylstra:
"it" is used to refer to a previously used *noun*, but "that" can refer to an *idea* that was just described.  (There are exceptions,  of course, but this is the general pattern.)


AriPerkkio (approved):
Good catch, thanks! 💯

